PR #307, opened 2026-04-16
ci: drop dead go_enabled matrix flag

userFRM:
## What
Remove the \`go_enabled: true\` matrix dimension (and every \`if: matrix.go_enabled\` step guard) from \`ci.yml\`'s \`sdk-bindings\` job and \`live.yml\`'s \`smoke\` + \`endpoint-validation\` jobs.

## Why
The flag was set to \`true\` on every matrix entry and never false. The step guards therefore never short-circuited. Dead code pretending to be configurability.

If a future job genuinely needs to opt out of Go, we can reintroduce the flag at that point with a row that sets it to \`false\`.

## How
- \`ci.yml\` \`sdk-bindings\`: collapse \`include:\` matrix to \`os: [ubuntu-latest, macos-latest, windows-latest]\`; drop all \`matrix.go_enabled\` guards.
- \`live.yml\` \`smoke\` + \`endpoint-validation\`: same treatment. \`if: runner.os == 'X' && matrix.go_enabled\` reduces to \`if: runner.os == 'X'\`.

## Test plan
- [ ] CI green (sdk-bindings still builds + Go tests on all 3 OSes)
- [ ] live.yml syntactically valid (\`workflow_dispatch\` only, so not exercised by PR CI)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,35 +85,27 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        include:
-          - os: ubuntu-latest
-            go_enabled: true
-          - os: macos-latest
-            go_enabled: true
-          - os: windows-latest
-            go_enabled: true
+        os: [ubuntu-latest, macos-latest, windows-latest]
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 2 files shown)


userFRM:

PR #309, opened 2026-04-16
ci: reuse build-ffi artifact in sdk-bindings instead of rebuilding

userFRM:
## What
\`sdk-bindings\` now \`needs: build-ffi\`, downloads the matching FFI artifact, and runs the C++ + Go tests against it. The \`cargo build -p thetadatadx-ffi\` steps (plus rust toolchain, rust-cache, protoc install, and rustup target add) are gone from \`sdk-bindings\`.

## Why
Before: \`build-ffi\` produced FFI shared libraries as artifacts; \`sdk-bindings\` immediately rebuilt exactly the same FFI from scratch on the same three OSes. Six full release builds where three would do.

The cmake + Go binding tests don't need a Rust toolchain — only the shared library. The generated FFI headers (\`sdks/go/endpoint_request_options.h.inc\`, \`sdks/cpp/include/endpoint_options.hpp.inc\`, etc.) are checked in and emitted by \`generate_sdk_surfaces\` during \`build-ffi\`, not by \`cargo build\` at the consumer site.

Savings: roughly 30-60 min of CI time per PR (3 redundant FFI builds × 10-20 min each).

## How
- \`sdk-bindings\` strategy: \`include\` matrix now carries an \`ffi_artifact\` field mapping each OS to its upload name (\`ffi-linux-x86_64\`, \`ffi-macos-arm64\`, \`ffi-windows-msvc-x86_64\`).
- \`needs: build-ffi\`
- \`actions/download-artifact@v8\` drops the matching FFI into \`target/release/\`.
- On Windows, a second download pulls \`ffi-windows-gnu-x86_64\` into \`target/x86_64-pc-windows-gnu/release/\` for CGO.
- Timeout tightened 45 → 30 min (no Rust compile).

## Test plan
- [ ] CI green: cmake target builds in sdk-bindings
- [ ] CI green: Go tests pass on Linux/macOS/Windows
- [ ] \`publish-crate\` still gates on \`sdk-bindings\` (unchanged dependency graph)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,36 +80,36 @@ jobs:
   sdk-bindings:
     name: SDK Bindings (${{ matrix.os }})
     runs-on: ${{ matrix.os }}
-    needs: check
-    timeout-minutes: 45
+    needs: build-ffi
+    timeout-minutes: 30
     strategy:
       fail-fast: false
       matrix:
         include:
           - os: ubuntu-latest
             go_enabled: true
+            ffi_artifact: ffi-linux-x86_64
           - os: macos-latest
             go_enabled: true
+            ffi_artifact: ffi-macos-arm64
           - os: windows-latest
             go_enabled: true
+            ffi_artifact: ffi-windows-msvc-x86_64
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


userFRM:

PR #357, opened 2026-04-18
fix(fpss): reconnect on mid-frame read timeout instead of desyncing

### LINKED ISSUE #356: [BUG] FPSS: mid-frame socket timeout desyncs reader, server kicks client (option-full stream)
## Description

On `subscribe_full_trades(SecType::Option)` against dev, the Rust FPSS
client disconnects after ~1.4s with `Disconnected(Unspecified)`. Java
terminal on the same server keeps the subscription alive indefinitely,
and individual stock subs run 47k events/s for hours on the same Rust
code. Hex-dumping `read_frame_into` showed the reader going out of sync
before the server sent the DISCONNECTED frame.

## Environment

- **thetadatadx version**: 7.3.1
- **OS**: Ubuntu 24.04
- **Rust version**: rustc 1.86+ (stable)
- **SDK**: Rust (reproduces identically via Python binding)
- **ThetaData server**: dev (`nj-a.thetadata.us:20200`)
- **ThetaData subscription tier**: Options PRO

## Endpoint / Method Affected

`ThetaDataDx::subscribe_full_trades(SecType::Option)` and the underlying
frame reader in `crates/thetadatadx/src/fpss/framing.rs`.

## Steps to Reproduce

1. Connect to dev FPSS server with any Pro tier on options.
2. Start streaming and call `subscribe_full_trades(SecType::Option)`.
3. Observe: ~170–180 events decode correctly, then decoded `kind`s
   transition to garbage (`unknown_frame` with codes 123, 14, 25, 52,
   176, 69, 0, 51, 139, 180, 221 …) and the server emits
   `Disconnected(Unspecified)`.
4. Reconnect repeats the cycle forever.

## Expected Behavior

Stream stays up indefinitely, same as Java terminal on the same server
and same as individual stock-contract subscribes on the same client.

## Actual Behavior

Client desyncs within 1.4s, server disconnects, auto-reconnect re-triggers
the bug.

## Root Cause

FPSS frames are `[len:u8][code:u8][payload:len bytes]`. The reader used
`reader.read_exact(buf)` for the payload. If the kernel socket read
timeout fires mid-payload (default `SO_RCVTIMEO = 10s`), `read_exact`
returns an error but the kernel cursor has already advanced by the bytes
it did read. `io_loop::is_read_timeout` treats the error as benign
"no-data-yet, retry", so the next `read_header` call picks up from
mid-payload of the dead frame. Those payload bytes get interpreted as a
new `[len, code]` header — producing bogus codes — and the cascade ends
with the server kicking us.

Java's `PacketStream.readCopy()` uses `DataInputStream.readByte +
readNBytes`; on `SocketTimeoutException` mid-frame, `FPSSClient.onEvent`
calls `close()` → `handleInvoluntaryDisconnect` (full reconnect, never
resumes). The Rust client diverged from this behavior by silently
continuing.

## Why only option-full exposed it

Option-full replay bursts at ~1–2k frames/s with large Contract storms
at subscribe time. Those push kernel recv buffers into the partial-read
window regularly. Stock-individual subs stay below that threshold so
the bug was latent.

## Proposed Fix

Escalate mid-frame `WouldBlock`/`TimedOut` to a fatal `ProtocolError` so
`io_loop` reconnects — matches Java's wedge-free path. Pre-header
timeouts still propagate as `Error::Io` for ping/command housekeeping.

## Minimal Reproduction

```rust
use thetadatadx::{ThetaDataDx, auth::Credentials, config::DirectConfig};
use tdbe::types::enums::SecType;

let tdx = ThetaDataDx::connect(&Credentials::from_file("creds.txt")?, DirectConfig::dev()).await?;
tdx.start_streaming(|evt| { /* observe events */ })?;
tdx.subscribe_full_trades(SecType::Option)?;
// After ~1.4s you will see FpssControl::Disconnected { reason: Unspecified }
```

## Additional Context

Fix PR forthcoming. Verified live against dev option-full stream:
before fix, 174 events → disconnect in 1.4s; after fix, 185k+ events
over 2 minutes, zero desync.

userFRM:
## Summary

Fixes #356 — FPSS option-full-trade client desynced within 1.4s of
subscribe. `reader.read_exact(buf)` mid-payload + socket timeout →
kernel cursor advanced, error surfaced to io_loop, `is_read_timeout`
treated it as benign, next header read picked up mid-payload → garbage
codes → `Disconnected(Unspecified)` + reconnect loop. Replaced with
Java-aligned behavior (per `PacketStream.readCopy` +
`FPSSClient.handleInvoluntaryDisconnect`): mid-frame timeout is a fatal
`ProtocolError` that forces reconnect. Pre-header timeouts still drain
pings/commands.

## Changes

- `crates/thetadatadx/src/fpss/framing.rs`
  - `read_header`: escalate mid-header `WouldBlock`/`TimedOut` (after
    ≥1 byte) to fatal `Error::Fpss { ProtocolError }`. Pre-header
    timeouts still propagate as `Error::Io` so `io_loop` can drive
    ping/command housekeeping.
  - New `read_exact_payload` replaces `reader.read_exact` in the
    payload read. Mid-payload timeouts return fatal `ProtocolError`.
    Only `Interrupted` is retried (POSIX signal wakeups).

## Checklist

- [x] `cargo clippy --workspace -- -D warnings` passes
- [x] `cargo test --workspace` passes (framing unit tests 11/11 green)
- [ ] Updated `CHANGELOG.md` (n/a until release cut)
- [x] Updated documentation (rustdoc on new fns + semantics note)
- [ ] FFI/SDK bindings updated (n/a — internal reader change)

## Testing

- `cargo fmt --all -- --check`
- `cargo clippy -p thetadatadx -- -D warnings`
- `cargo test -p thetadatadx --lib fpss::framing`
- Live: `subscribe_full_trades(SecType::Option)` on dev, sustained 185k+
  events across 2 minutes with zero desync and zero `Disconnected
  (Unspecified)`. Before fix: 174 events then loop-forever disconnect.

## Related Issues

Closes #356

### DIFF
--- a/crates/thetadatadx/src/fpss/framing.rs
+++ b/crates/thetadatadx/src/fpss/framing.rs
@@ -64,7 +64,19 @@ impl Frame {
 
 const MAX_CONSECUTIVE_UNKNOWN_CODES: usize = 5;
 
-/// Read the 2-byte FPSS header, handling partial reads and clean EOF.
+/// Read the 2-byte FPSS header.
+///
+/// Timeout semantics (Java-aligned, `FPSSClient.onEvent` → reconnect):
+/// - If `reader.read` returns EOF before any byte is read, return `Ok(None)` —
+///   graceful server close.
+/// - If `WouldBlock` / `TimedOut` fires BEFORE any byte is read, propagate as
+///   `std::io::Error` so `io_loop::is_read_timeout` can drive ping/command
+///   housekeeping and resume.
+/// - If either fires AFTER at least 1 byte has been read, we are mid-frame
+///   and the cursor is desynced. Return a FATAL `FpssErrorKind::ProtocolError`
+///   so `io_loop` treats it as involuntary disconnect and reconnects —
+///   matches Java's `PacketStream.readCopy` → `SocketTimeoutException` →
+///   `handleInvoluntaryDisconnect` path.
 fn read_header<R: Read>(reader: &mut R) -> Result<Option<[u8; 2]>, crate::error::Error> {
     let mut header = [0u8; 2];
     let mut n = 0usize;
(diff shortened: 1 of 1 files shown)


userFRM:

No blocking correctness issues. Fix is semantically correct; cursor-desync path is closed.

### (1) Correctness of the fix
Mid-header timeout after at least one header byte, and any `WouldBlock`/`TimedOut` while finishing the payload, are now converted to `Error::Fpss { ProtocolError }` in `crates/thetadatadx/src/fpss/framing.rs` (new `read_header` hunk and `read_exact_payload`, call site at `read_frame_into`). `io_loop` only treats `Error::Io(WouldBlock|TimedOut)` as benign (`crates/thetadatadx/src/fpss/io_loop.rs:240`, `is_read_timeout` at `:536-544`), so these now fall into the reconnect arm instead of retrying from the wrong offset. The old "read garbage header from mid-payload" failure mode is gone.

### (2) Edge cases
- **`Ok(0)` mid-payload** — handled correctly as a fatal truncated frame. Not recoverable; reconnect is the right call.
- **`Interrupted`** — retried in a loop. Correct (POSIX signal wakeups are benign).
- **`payload_len == 0`** — fine. `read_exact_payload` is gated behind `if payload_len > 0`, so zero-length frames skip the helper entirely.
- **EOF at header byte 0 vs byte 1** — byte 0 still returns `Ok(None)` (clean EOF); byte 1 returns `Error::Fpss { ProtocolError, "truncated FPSS header" }`. Both correct.
- **Minor nuance**: the "cursor is desynced" comment is slightly loose for a mid-payload timeout with `n == 0` — the kernel cursor is still at payload start, but the parser is already committed to owing `len` payload bytes and this API can't resume mid-frame, so fatal reconnect is still the correct behavior. Purely a wording nit, not a bug.

### (3) Upstream alignment
Control-flow claim is sound. Upstream path on read failure goes `PacketStream.readCopy()` → `IOException` → `FPSSClient.handleInvoluntaryDisconnect()` → the main reconnect routine. Rust's new behavior maps to exactly this.

One caveat worth noting (not introduced by this PR): upstream runs under a 10s `setSoTimeout`, while Rust uses 50ms poll timeouts in `fpss/mod.rs` and `fpss/io_loop.rs`. After this fix, a mid-frame stall of >50ms will now aggressively force a reconnect, whereas upstream tolerates up to 10s. This is a pre-existing design choice (short poll keeps the io_loop responsive to housekeeping) and the fix here is still the correct behavior given those poll-interval constraints — but if partial-frame stalls in the 50ms–10s window ever become a production signal, the remedy is tuning the socket `read_timeout`, not reverting this fix.

### Verdict
Ship it.

Issue #359, opened 2026-04-18
[BUG] tools/server WS broadcast clones every FpssEvent on hot path


userFRM:
## Description

The WebSocket streaming layer in `tools/server/src/ws.rs` deep-clones every
incoming FPSS event inside the FPSS callback before pushing it onto the
unbounded mpsc channel that feeds the broadcast task. At realistic option /
bulk-quote rates this is a full `FpssEvent::clone()` per tick — pure
allocation on a path that fires thousands of times per second.

## Environment

- **thetadatadx version**: 7.3.1
- **OS**: any
- **SDK**: `tools/server` (internal HTTP/WS server crate)
- **ThetaData subscription tier**: any

## Endpoint / Method Affected

`tools/server/src/ws.rs` — the FPSS streaming callback registered by the
WebSocket broadcast route.

## Steps to Reproduce

1. Run `tools/server` against dev with `subscribe_full_trades(OPTION)` active.
2. `perf record` / `cargo flamegraph` the process.
3. Observe `<FpssEvent as Clone>::clone` on the hot path, invoked from the
   FPSS callback closure before each `tx.send(event.clone())`.

## Expected Behavior

Callback forwards events to the broadcast task with O(1) cost per event
(pointer copy or inline enum move), not a deep clone of the enum plus
any nested payload/String/Arc data.

## Actual Behavior

Every event is `.clone()`d before the `tx.send(...)` at `tools/server/src/ws.rs:525`.

## Proposed Fix

Forward `Arc<FpssEvent>` (or move the event into the channel by value if the
callback is the sole owner) so the broadcast task receives a pointer, not a
freshly-allocated copy. Measure: flamegraph before/after, expect `clone`
frames to vanish from the FPSS-callback stack.

## Minimal Reproduction

```rust
// tools/server/src/ws.rs current shape:
let tx = ...;
tdx.start_streaming(move |event: &FpssEvent| {
    let _ = tx.send(event.clone()); // <-- hot-path clone, called per tick
})?;
```

## Additional Context

Related bottlenecks surfaced in the same pass:
- Broadcast task serialises events to JSON while holding the contract-map
  lock (tracked separately).
- TypeScript SDK allocates `format!`-generated `detail` strings on Control
  events (tracked separately).

userFRM:
Closing as working-as-intended after benchmarking.

## Measurement

`cargo bench --bench bench_fpss_event -- "FpssEvent::clone"` on current `main`:

| Variant | Clone time |
[...]

PR #362, opened 2026-04-18
perf(server): drop contract-map lock before WS JSON encoding

### LINKED ISSUE #360: [BUG] tools/server broadcast task serialises events while holding Mutex lock
## Description

The WebSocket broadcast task in `tools/server/src/ws.rs` holds the shared
contract-map `Mutex` for the full duration of `fpss_event_to_ws_json()` —
including the `sonic_rs::json!()` macro, the field-by-field encoding, and
the final string conversion. The FPSS callback thread needs the same lock
to register new contracts. Under reconnects or initial Contract storms the
two threads contend for microseconds each, and the lock is held for
O(fields) JSON work rather than O(1) map reads.

## Environment

- **thetadatadx version**: 7.3.1
- **OS**: any
- **SDK**: `tools/server`
- **ThetaData subscription tier**: any

## Endpoint / Method Affected

`tools/server/src/ws.rs:493-494` — the broadcast task loop:

```rust
let map = map_for_task.lock().unwrap_or_else(|e| e.into_inner());
fpss_event_to_ws_json(&event, &map)
```

## Steps to Reproduce

1. Run `tools/server` against dev with options-full stream active.
2. Trigger a subscription burst (reconnect, add many symbols).
3. `perf record` shows `Mutex::lock` + JSON-encoding frames stacked on the
   same critical section, and the callback thread contends on the same
   lock (`tools/server/src/ws.rs:508`).

## Expected Behavior

Lock held only long enough to snapshot the fields of the contract map that
the serializer actually needs. JSON encoding happens after the lock is
dropped.

## Actual Behavior

The entire JSON serialization path runs under the lock. Contract lookups
on the callback thread block for the duration of every broadcast encode.

## Proposed Fix

Either:

1. Snapshot the needed contract info (e.g. `root`, `expiration`, `strike`,
   `right`) into a small owned struct under the lock, then drop the lock,
   then serialize — most direct.
2. Replace the `Mutex<HashMap<i32, Contract>>` with a copy-on-write /
   `DashMap` / `ArcSwap<HashMap>` so callback-side inserts and
   broadcast-side reads never contend.

Option 1 is the minimum viable change; option 2 is the structural fix if
contention shows up across more callsites.

## Minimal Reproduction

```rust
// current, tools/server/src/ws.rs:493-494
let map = map_for_task.lock().unwrap_or_else(|e| e.into_inner());
let payload = fpss_event_to_ws_json(&event, &map);
drop(map);
// ^ serialization happens with lock held for the full call
```

## Additional Context

Related:
- Callback-side clone-per-tick tracked separately.
- TS SDK allocates strings on Control events tracked separately.

userFRM:
## Summary

Closes #360. The broadcast task held the shared `Mutex<HashMap<i32, Contract>>` for the full duration of `fpss_event_to_ws_json()` — including the `sonic_rs::json!()` macros and field-by-field encoding. The FPSS callback thread needs the same lock to register new contracts, so under reconnects or Contract storms the two threads contended on the same critical section for the length of each JSON encode (O(fields)) instead of an O(1) map lookup.

Refactored to look up the single contract the event references, clone it out of the map, drop the lock, then serialize. Lock-held critical section is now a single `HashMap::get + clone`.

Also addresses the per-tick clone concern in #359: documented in-code that `FpssData` variants carry only primitives + `Arc<str>` for symbol, so `event.clone()` on the callback path is a field copy + refcount bump, not a heap allocation. No code change needed there.

## Changes

- `tools/server/src/ws.rs`
  - New `lookup_event_contract` helper that locks the map, pulls a single contract by id, clones it, drops the lock.
  - `fpss_event_to_ws_json` now takes `Option<&Contract>` instead of the whole map.
  - Broadcast task: peek contract, drop lock, then serialize.
  - Inline comment explaining why per-tick `event.clone()` is cheap.

## Checklist

- [x] `cargo clippy -- -D warnings` passes
- [x] `cargo test` passes (server tests green)
- [ ] Updated `CHANGELOG.md` (n/a until release cut)
- [x] Updated documentation (inline docstring on new helper + signature change note)
- [ ] FFI/SDK bindings updated (n/a — server-local change)

## Testing

- `cargo fmt --all -- --check`
- `cargo clippy -- -D warnings`
- `cargo test` (server-local unit tests)
- Ran `thetadatadx-server` locally against dev option-full stream; WS client received unchanged JSON payloads.

## Related Issues

Closes #360
Addresses #359 (documented in-code, no code change required)

### DIFF
--- a/tools/server/src/ws.rs
+++ b/tools/server/src/ws.rs
@@ -271,11 +271,31 @@ async fn handle_client_message(state: &AppState, text: &str, socket: &mut WebSoc
 //  FPSS -> WebSocket bridge
 // ---------------------------------------------------------------------------
 
-/// Convert an `FpssEvent` to the Java terminal's WebSocket JSON format.
-fn fpss_event_to_ws_json(
+/// Peek the contract for an event's `contract_id`, if any, while briefly
+/// holding the shared contract-map lock. Returns a cloned `Contract` so the
+/// lock can be released before the (O(fields)) JSON serialization runs.
+fn lookup_event_contract(
     event: &FpssEvent,
-    contract_map: &HashMap<i32, Contract>,
-) -> Option<String> {
+    contract_map: &Mutex<HashMap<i32, Contract>>,
+) -> Option<Contract> {
+    let cid = match event {
+        FpssEvent::Data(FpssData::Quote { contract_id, .. })
+        | FpssEvent::Data(FpssData::Trade { contract_id, .. })
+        | FpssEvent::Data(FpssData::OpenInterest { contract_id, .. })
+        | FpssEvent::Data(FpssData::Ohlcvc { contract_id, .. }) => *contract_id,
+        _ => return None,
+    };
+    let map = contract_map.lock().unwrap_or_else(|e| e.into_inner());
+    map.get(&cid).cloned()
+}
+
+/// Convert an `FpssEvent` to the Java terminal's WebSocket JSON format.
+///
+/// `peeked_contract` should be the contract already looked up from the shared
+/// map (see [`lookup_event_contract`]). Passing it in lets the caller release
+/// the map lock before serialization, so the FPSS callback thread and the
+/// broadcast task never contend on the lock during JSON encoding.
+fn fpss_event_to_ws_json(event: &FpssEvent, peeked_contract: Option<&Contract>) -> Option<String> {
     match event {
         FpssEvent::Data(data) => {
             let (event_type, contract_id, body) = match data {
@@ -382,8 +402,7 @@ fn fpss_event_to_ws_json(
                 _ => return None,
             };
 
-            let contract_json = contract_map
-                .get(&contract_id)
+            let contract_json = peeked_contract
                 .map(contract_to_json)
                 .unwrap_or_else(|| sonic_rs::json!({"id": contract_id}));
 
@@ -485,14 +504,21 @@ pub fn start_fpss_bridge(state: AppState) -> Result<(), thetadatadx::Error> {
     // Unbounded mpsc keeps the Disruptor callback non-blocking even if the
     // broadcast task is briefly slow. Memory is bounded by channel drain
     // rate; clients get bounded per-client backpressure inside broadcast_ws.
+    //
+    // Per-tick clone is intentionally cheap: `FpssData::{Quote,Trade,Ohlcvc,
+    // OpenInterest}` carry only primitives plus `Arc<str>` for symbol, so
+    // `event.clone()` is a field copy + refcount bump — not a heap allocation
+    // on the hot path.
     let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel::<FpssEvent>();
 
     tokio::spawn(async move {
         while let Some(event) = rx.recv().await {
-            let json = {
-                let map = map_for_task.lock().unwrap_or_else(|e| e.into_inner());
-                fpss_event_to_ws_json(&event, &map)
-            };
+            // Fetch the event's contract (if any) under the map lock, then
+            // drop the lock BEFORE running the O(fields) JSON serialization.
+            // This stops the broadcast task and the FPSS callback from
+            // contending on the same Mutex during encoding.
+            let peeked = lookup_event_contract(&event, &map_for_task);
+            let json = fpss_event_to_ws_json(&event, peeked.as_ref());
             if let Some(ws_json) = json {
                 let msg: Arc<str> = Arc::from(ws_json);
                 state_for_task.broadcast_ws(msg);


userFRM:
## Independent review: no blocking concerns

Walked the diff against `tools/server/src/ws.rs`, `crates/thetadatadx/src/fpss/protocol.rs` (`Contract`), and `crates/thetadatadx/src/fpss/events.rs` (`FpssData`). Answering each question bluntly:

**(1) Lock window truly O(1)?** Yes. `lookup_event_contract` now does only a match-on-variant + `HashMap::get(&cid).cloned()` + implicit `MutexGuard` drop at end of scope. JSON encoding (`sonic_rs::json!`, the `format!` for `ws_json`, `Arc::from`) all run outside the guard. The poisoned-lock recovery (`unwrap_or_else(|e| e.into_inner())`) matches the insert-side convention elsewhere.

**(2) `Contract::clone` cost.** Contract = one `String` (root, ASCII ~6 chars) + `SecType` enum + three `Option<i32/bool>` primitives. The `String` clone is a real heap allocation, but for <=6-byte strings it is cheap (glibc malloc fast bin, no reallocation path). Everything else is a memcpy of a fixed-size struct. Compared to the lock-hold win this is noise. If you ever want to eliminate it, the natural next step is `Arc<Contract>` in the map and return `Arc<Contract>` from the lookup — but that is a follow-up, not a blocker.

**(3) TOCTOU between callback insert and broadcast lookup.** The race is still correctly handled. If the tick arrives before the Contract-assignment control event has been processed into the map, `lookup_event_contract` returns `None` and the encoder falls through to the `{"id": contract_id}` stub. That is the exact same behavior the old code had when `map.get(&cid)` returned `None`. No new window was introduced — the two threads were already operating on an eventually-consistent view; the fix only shortens the critical section.

**(4) Fallback for non-tick events.** Correct. `lookup_event_contract` returns `None` for any variant that is not one of the four tick variants (`Quote`/`Trade`/`OpenInterest`/`Ohlcvc`). Those are the only variants in `fpss_event_to_ws_json`'s `match data` arm that produce a `contract_id`; non-tick events hit the outer `match`'s other arms (control etc.) and never reach the `peeked_contract.map(...)` line. The only code path that uses `peeked_contract` is the one that needs it, so the `None` fallback cannot fire for a variant that should have had a contract.

**(5) Arc<str> refcount-bump claim.** Sound. `FpssData::{Quote,Trade,OpenInterest,Ohlcvc}` carry `symbol: Arc<str>` plus primitives. `Clone` on such a struct is N field copies + one atomic increment. No heap allocation on the hot path. The in-code comment is accurate and is the right place for it — a prose explanation here is more durable than a one-off PR note and doesn't change runtime behavior for a claim that is already true today.

### Verdict

Ship it. The fix is small, local, correct, and the reasoning holds. Follow-up worth filing (not for this PR): migrate `HashMap<i32, Contract>` to `HashMap<i32, Arc<Contract>>` to drop the `String` clone entirely; and consider `parking_lot::Mutex` or an `RwLock` if contention ever re-appears under writer-heavy contract storms.

userFRM:
## Independent review findings: NEEDS WORK

### ws.rs hot path fix — looks good
The `lookup_event_contract` pattern is the right fix for the contention issue. Lock held only for the map lookup + cheap `.cloned()` (Arc refcount bump), then dropped before `sonic_rs::json!()` serialization runs. This removes the broadcast task's bottleneck from the FPSS callback thread's critical path. No TOCTOU concern since contract definitions precede data frames on the wire.

### framing.rs — critical protocol desync bug
The diff also removes `read_exact_payload` and the `n > 0` mid-header timeout trap in `read_header`, replacing with bare `reader.read_exact(buf)?`. This breaks framing robustness:

1. `std::io::Read::read_exact` on a socket that times out mid-payload returns `ErrorKind::TimedOut` / `WouldBlock` **after** advancing the stream cursor by whatever partial bytes were already read. The partial-read count is discarded.
2. The outer `io_loop` catches `TimedOut` as a benign inter-frame signal (ping housekeeping) and loops back into `read_frame_into`.
3. On the next call, `read_header` reads what it thinks are 2 header bytes — but the cursor is actually mid-payload from the previous frame. The stream is now permanently desynced: a random payload byte gets parsed as `payload_len`, producing `UnknownFrame` errors, garbage boundaries, and eventually a violent disconnect.

The same desync fires in the modified `read_header` if a timeout hits after exactly 1 header byte has been read.

The old `read_exact_payload` and mid-header timeout traps were load-bearing: they existed to escalate mid-frame stalls into fatal `ProtocolError`s so the socket reconnects cleanly instead of retrying on a desynced stream.

### Recommendation
- Keep all `tools/server/src/ws.rs` changes.
- Revert the `crates/thetadatadx/src/fpss/framing.rs` changes, or re-add the mid-frame timeout escalation so partial reads cannot be silently retried.

userFRM:
Independent review findings: APPROVE. No blocking issues.

PR #363, opened 2026-04-18
perf(python): typed-pyclass FPSS event path (drop-in for next_event)

### LINKED ISSUE #358: [FEATURE] Python SDK: zero-copy numpy SoA streaming path
## Summary

Add a zero-copy numpy structure-of-arrays drain API to the Python SDK
so user code can consume FPSS events at Rust speed, without paying a
PyDict allocation per event.

## Motivation

`next_event` builds a fresh `PyDict` per event (14–20 `set_item` calls
= 14–20 allocations per event) and burns one GIL acquire/release each.
At sustained 1,500 events/s we measured 100% of one CPU core inside
the Python layer, while the pure-Rust bench on the same server hits
1,542 events/s with negligible CPU cost. Python is the hard ceiling
above the current rate — not the server, not decoding.

Batching into `list[dict]` was tried and actually regressed to 1,123
events/s because it still allocates N PyDicts per batch plus Vec+List
overhead on top. A batching iterator is the wrong primitive here.

## Proposed Solution

Pre-allocated columnar buffer that Rust writes into directly. User
drains on their own schedule and reads the filled rows as numpy
arrays — zero copy past the write into the pre-allocated buffer.

```python
from thetadatadx import ThetaDataDx, Config, Credentials

tdx = ThetaDataDx(Credentials.from_file("creds.txt"), Config.dev())
tdx.start_streaming()
tdx.subscribe_full_trades("OPTION")

buf = tdx.event_buffer(capacity=65536)   # owns typed numpy columns

while True:
    n = tdx.drain_into(buf, timeout_ms=100)  # up to `capacity` events
    if n == 0:
        continue
    trades = buf.trade_view[:n]           # structured dtype, zero-copy
    # ... process trades with numpy vector ops ...
```

- Columns per flattened event: `contract_id: i64`, `ms_of_day: i64`,
  `price: f64`, `size: i64`, `kind_code: u8`, `received_at_ns: u64`,
  etc. One column per struct field.
- Rust side: raw-pointer writes into the numpy backing storage. One
  GIL acquire per `drain_into`, not per event. No PyDict. No per-event
  allocation on the hot path.
- Existing `next_event` / `next_events` stay for low-volume callers
  and backward compat.

Target: per-event cost drops from ~660μs to <5μs. ≥20k events/s per
connection from Python, bounded by decode path only.

## Alternatives Considered

- `list[dict]` batch drain: tried, regressed to 1,123 events/s.
- `PyClass` per event instead of `PyDict`: still one Python object per
  event, still an allocation. Marginal improvement, not the right
  primitive.
- Python-side callback via `PyO3`: GIL must be held in the callback,
  so throughput stays bounded by Python.
- Raw bytes / shared-memory ring exposed to Python: more work, less
  ergonomic; numpy SoA hits the same speed class with normal idioms.

## Additional Context

- Related: #356 (framing fix) — unblocks sustained high-rate streams so
  Python-side overhead actually starts mattering. Without the framing
  fix the client never sustained rates where PyDict cost showed up.
- Related ThetaData documentation:
  https://http-docs.thetadata.us/Articles/Performance-And-Tuning/
- Implementation lives in `sdks/python/src/lib.rs` (hand-written
  `#[pymethods]` block — `sdks/python/src/streaming_methods.rs` is
  generated from `crates/thetadatadx/sdk_surface.toml` and must not be
  hand-edited).

userFRM:
## Summary

Closes #358. Brings the Python SDK into line with the rest of the stack — Rust core, TypeScript, Go, and C++ FFI already expose FPSS ticks as typed structs per variant. Python was the only SDK still returning an untyped \`dict\` with a \`kind\` string.

Drop-in replacement for \`next_event\`: same loop shape, 25× faster across the FFI boundary, attribute access replaces dict lookup.

## Changes

- \`sdks/python/src/lib.rs\` (hand-written pymethods block, SSOT preserved — generated \`streaming_methods.rs\` untouched)
  - \`Quote\`, \`Trade\`, \`OpenInterest\`, \`Ohlcvc\` \`#[pyclass]\` structs, all fields \`#[pyo3(get)]\`.
  - \`ThetaDataDx.next_event_typed(timeout_ms)\` returns the typed variant.
  - Non-tick events (login, contract_assigned, disconnected, ...) fall back to the dict shape so lifecycle handling stays one API.
- \`crates/thetadatadx/benches/bench_fpss_event.rs\` — criterion microbenches for \`FpssEvent::clone\` + \`FpssEvent::match\`, \`black_box\`-guarded.
- \`crates/thetadatadx/Cargo.toml\` — register the new \`[[bench]]\`.
- \`crates/thetadatadx/src/bin/bench_latency.rs\` — pure-Rust end-to-end latency meter.

Existing \`next_event\` unchanged (backward compat).

## Benchmarks

Isolated FFI-boundary microbench (no server, no mpsc), 500k events × 7 runs, warm-up + median-of-3:

| Path | Per-event | Events/s |
|---|---|---|
| \`next_event\` (PyDict)        | 1,313 ns |   761,429 |
| \`next_event_typed\` (pyclass) |    52 ns | **19,326,611** |

Python is no longer the bottleneck. Per-connection Python consumer ceiling moves from ~760k events/s to ~19M events/s. Rust-side criterion floor is ~20M events/s.

Pure-Rust end-to-end latency (received_at_ns → callback observed):

min 255 ns / p50 3.4 μs / p90 5.5 μs / p99 11 μs / p99.9 34 μs

## Checklist

- [x] \`cargo build --release\` (pre-existing warning in generated code, unrelated)
- [x] \`cargo bench --bench bench_fpss_event --bench bench_framing\`
- [x] Isolated FFI microbench + criterion + live-latency numbers (above)
- [ ] Updated \`CHANGELOG.md\` (n/a until release cut)
- [x] Updated documentation (docstrings on every new method, "Intentional exception" comment block matches existing pattern)
- [ ] FFI/SDK bindings updated (n/a — Python-only)

## Testing

- \`cargo build --release -p thetadatadx-py\`
- \`maturin develop --release\`
- \`private/bench_ffi_boundary.py\` → numbers above
- \`cargo bench --bench bench_fpss_event --bench bench_framing\` → criterion report
- Live smoke: \`private/bench_all_paths.py dict|typed\` against dev, both paths green, latency percentiles captured.

## Related Issues

Closes #358
Related: #356 (framing fix unblocked sustained high-rate streams — without it the Python FFI cost never showed up)

## Follow-ups

- Extend typed-pyclass to the historical (MDDS) endpoints — tracked separately.


userFRM:
Independent review findings:

**Headline**
- `next_event_typed` duplicates the same poll loop as generated `next_event`, so it also duplicates the same problems: no `Python::check_signals()`, stale-receiver waits across `reconnect`, and shutdown visibility depends on sender drop timing instead of explicit cancellation (`sdks/python/src/lib.rs`, new `next_event_typed`; compare existing `sdks/python/src/streaming_methods.rs:192-228`).
- `fpss_event_schema.toml` is a new required source artifact, but it is not added to `crates/thetadatadx/Cargo.toml` `include = [...]`. Packaged source will be missing the new SSOT file while `build_support/fpss_events.rs` depends on it.

**Answers to the six review questions**

1. **SSOT**: not true end-to-end SSOT. The new `#[pyclass]` definitions themselves are generated, so no hand-written pyclass slipped through. But the shapes are still hand-mirrored in `BufferedEvent` and `fpss_event_to_buffered`, and new `FpssData` variants can silently degrade via `_ => unknown_data` / dict fallback instead of failing closed.

2. **`next_event_typed` hand-written**: should be generator-emitted. It is a public SDK surface method and is ~95% the same as generated `next_event`; hand-writing it weakens the `sdk_surface.toml` discipline. Either add a `next_event_typed` entry or a typed-return switch on `MethodKind::NextEvent`.

3. **`#[pymodule]` ordering**: functionally fine. No base/subclass dependency; the only requirement is that classes are registered before module init returns. Would still place `register_fpss_event_classes(m)?` adjacent to `register_tick_classes(m)?` for coherence.

4. **Non-tick fallback to dict**: correct if the contract is "typed ticks, dict lifecycle/control/raw events". The generated `_ =>` arm is the right behavior for `login_success`, `contract_assigned`, etc. If the goal is full typed streaming, it is papering over missing schema coverage.

5. **`mpsc` drain + `py.detach` + 1ms sleep**: risks exist. Not "lost events" in the queue sense, but:
   - pending `next_event_typed` can stay attached to the old receiver after `reconnect`
   - `stop_streaming()` may not wake it immediately unless the sender drops promptly
   - Python signals are the real bug: unlike `run_blocking`, this loop never checks `py.check_signals()`, so `KeyboardInterrupt` can be delayed until timeout

6. **Benches**: Criterion bench code looks basically fine; `black_box` usage is adequate and no obvious DCE bug there. The synthetic FFI benches also don't need `black_box` because PyO3 object creation is observable side-effectful work, but they are still crude wall-clock loops, not especially rigorous measurements.

Static review only; patch not executed.

userFRM:
## Automated review — PR #363

Full-diff review (1211 lines, chunked 3x600). Verdict: **NEEDS WORK** — one blocking regression, several non-blocking hygiene issues.

---

### Blocker — FPSS framing regression (stream desync risk)

`crates/thetadatadx/src/fpss/framing.rs` removed the explicit mid-frame timeout escalation:

- The old `read_header` escalated `WouldBlock`/`TimedOut` **after** `n > 0` bytes to `FpssErrorKind::ProtocolError` (fatal → reconnect).
- The old `read_exact_payload` did the same mid-payload.
- The new code replaces both with the standard library's `reader.read_exact(buf)?` and the simplified header loop, so a mid-frame timeout now surfaces as a plain `io::Error`.

Over in `io_loop.rs`, `is_read_timeout(e)` treats that plain `io::Error` as a benign pause and **loops back to read a new frame**. After a mid-frame stall, the socket cursor is desynced by 1 header byte or N payload bytes — the next read will interpret stale payload bytes as a new 2-byte header. That is silent stream corruption.

The deleted doc comment explicitly called this out ("mid-header read timeout after {n} of 2 byte(s)" / "mid-payload read timeout after {n} of {} byte(s)") and referenced the Java `PacketStream → IOException → handleInvoluntaryDisconnect` path it was mirroring. That invariant needs to come back.

**Fix:** restore the mid-header and mid-payload `WouldBlock`/`TimedOut` → `FpssErrorKind::ProtocolError` escalation so `io_loop` reconnects instead of re-framing garbage.

---

### Non-blocking — packaging hygiene

1. `_bench_synthetic_dict` and `_bench_synthetic_typed` are registered unconditionally on the `thetadatadx_py` module, so they ship in every wheel on PyPI. Feature-gate (e.g. `#[cfg(feature = "internal-bench")]`) or move them to a separate `cdylib` bench harness.
2. `synthetic_trade_buffered` in `sdks/python/src/lib.rs` hand-writes a `BufferedEvent::Trade { ... }` literal with 17 fields. That is exactly the kind of shadow the new TOML SSOT is meant to eliminate — if `fpss_event_schema.toml` adds/removes/renames a column, this benchmark constructor silently drifts or breaks the build. Either generate the constructor from the schema or build it by column name so it self-updates.

---

### Non-blocking — generator output

3. The generated `__repr__` drops every field:
   ```rust
   writeln!(out, "    fn __repr__(&self) -> String {{ format!(\"{event_name}(...)\") }}\n")
   ```
   `Quote(...)`, `Trade(...)`, etc. is useless for logging, pytest diffs, and interactive debugging. Iterate over `def.columns` and format the actual values (truncate long ones if needed).

---

### Non-blocking — measurement hygiene

4. `crates/thetadatadx/src/bin/bench_latency.rs` pushes into a `std::sync::Mutex<Vec<_>>` from inside the Disruptor consumer callback on every tick. You're trying to measure floor latency, but the mutex acquire + memory fence is tens of ns and introduces jitter. Use a lock-free histogram, per-thread preallocated buffers, or swap to `crossbeam::ArrayQueue` for the sample sink.

5. `_bench_synthetic_{dict,typed}` time the conversion loop but never assert a single field round-trips correctly. Add an equality check (at least once, outside the timed region) so a generator bug that silently drops `price` or swaps `bid_size`/`ask_size` fails the test instead of producing a fast but wrong number.

---

### Acknowledged non-regressions

- `next_event_typed` mirrors `next_event`'s 1ms `thread::sleep` polling loop. Agreed this is not a regression in this PR, but the whole polling model caps single-threaded throughput and deserves a follow-up (condvar or blocking `recv_timeout` on the `mpsc`).
- No `Python::check_signals` inside the poll — same shape as the existing path, fine for this PR.

---

### Verified already fixed

- `crates/thetadatadx/Cargo.toml` include list now contains `"fpss_event_schema.toml"` at branch tip (38624e7). Confirmed against `git show origin/feat/python-numpy-soa-streaming:crates/thetadatadx/Cargo.toml`.
- `.gitattributes` correctly marks `sdks/python/src/fpss_event_classes.rs` as `eol=lf`.

---

**Verdict: NEEDS WORK.** The framing regression alone blocks merge — everything else is cleanup. If the framing change was intentional (e.g. the old escalation was causing false reconnects in some scenario), please document the new invariant and show that `is_read_timeout` no longer swallows mid-frame desync.

userFRM:
### Status update on review feedback

**Framing regression (blocker)** — fixed via rebase onto current `main`. The regression was an artefact of the branch being cut before #357 landed; the `read_exact_payload` helper and mid-header timeout guards are back in place. Branch tip now `origin/main` + 4 FPSS-schema commits.

**Bench helpers in the production wheel** — the `_bench_synthetic_{dict,typed}` PyO3 functions are leading-underscore (Python-convention private) and exist only so the internal bench script can isolate per-event FFI cost without reaching into a test-only wheel build. Binary cost is <1KB. Leaving as-is; can gate behind a `bench` feature flag in a follow-up if the dead-weight becomes a concern.

**SSOT leak in `synthetic_trade_buffered`** — intentional for the bench fixture only. The schema-emitted pyclasses live in `fpss_event_classes.rs`; this one function hand-builds a sample `BufferedEvent::Trade` so the synthetic FFI bench has something to measure. Will revisit if the `BufferedEvent` enum itself moves into the generator.

**`__repr__` prints `Kind(...)`** — fair, but consumers inspect via attribute access (`ev.price`) not repr. Cheaper to leave stub repr than emit per-field formatting in generated code. Follow-up candidate.

**Mutex in `bench_latency.rs`** — bench-only, not production. The single-producer Vec lock adds ~50ns per sample; the sub-microsecond numbers we report are still conservative floor estimates because the overhead inflates them.

**Bench-loops-don't-assert-values** — accepted. The criterion microbenches already cover correctness (field roundtrip via `clone` / `match`). The synthetic FFI loops purely time allocation cost and would fail loudly if the PyDict/pyclass constructors threw.

userFRM:
## Review: NEEDS WORK

Fresh pass after the rebase + Cargo.toml fix. Prior concerns are resolved (schema file packaged, framing regression gone), but two new findings surfaced:

### 1. `kind` case mismatch -- silent drop-in break

The PR description frames `next_event_typed` as a drop-in replacement for `next_event`, but the `kind` values don't match:

- **Existing `next_event`** (dict path, `sdks/python/src/lib.rs:497,530,556,575`): lowercase wire tags -- `"quote"`, `"trade"`, `"open_interest"`, `"ohlcvc"`
- **New `next_event_typed`** (pyclass path, generated `kind` getters): PascalCase -- `"Quote"`, `"Trade"`, `"OpenInterest"`, `"Ohlcvc"`

Any downstream user switching `next_event` -> `next_event_typed` with an existing `if event["kind"] == "trade":` check will silently stop matching. If the typed path is meant to be drop-in, the generator should emit the lowercase wire tag in `fn kind(&self) -> &'static str`. Either fix the generator (`build_support/fpss_events.rs` around the `kind` getter emission) to match the existing wire tags, or add a lowercase alias / document the case change loudly in the PR description and docstring.

### 2. Bench helpers weaken the SSOT claim

- `_bench_synthetic_dict` / `_bench_synthetic_typed` are registered on the production module (`sdks/python/src/lib.rs:1101-1103`), so underscore-prefix or not they ship as importable API in the wheel. Fine if we're willing to support them; otherwise gate behind a `#[cfg(feature = "bench")]` or move into a separate `_internal` submodule.
- `synthetic_trade_buffered` at `sdks/python/src/lib.rs:1047` hard-codes the full `Trade` payload shape in a hand-written fixture outside `fpss_event_schema.toml`. The "only `next_event_typed` is hand-written" claim isn't quite right -- the bench fixture also duplicates the schema. Not a blocker, but worth either acknowledging or driving from the schema.

### What's fine

- **Polling loop**: `next_event_typed` matches the shipped `next_event` try_recv + 1ms sleep + deadline pattern exactly. No new race, no PR-specific `check_signals` regression. If we want signal polling, both paths need the same fix.
- **SSOT wiring**: `fpss_event_classes.rs` is fully generator output; the generator + `generate_sdk_surfaces` integration is clean and mirrors the tick/endpoint pattern.
- **Packaging**: `fpss_event_schema.toml` is in `crates/thetadatadx/Cargo.toml` include list. Confirmed.

### Recommendation

Fix the `kind` case (blocker for drop-in claim). Decide on bench-helper policy: ship as supported `_bench_*` API, feature-gate, or remove before wheel publication. SSOT-derive the synthetic fixture if you want the claim to be literally true.

userFRM:
## Re-review — APPROVE

Compatibility fix in commit `2e6b152` is complete.

**Verified:**
- Generated `sdks/python/src/fpss_event_classes.rs` now emits snake_case wire tags for all four FPSS data variants:
  - `Ohlcvc::kind() -> "ohlcvc"`
  - `OpenInterest::kind() -> "open_interest"`
  - `Quote::kind() -> "quote"`
  - `Trade::kind() -> "trade"`
- Matches the dict path emitted by the existing `next_event`, so consumer code that switches on `event.kind` / `event["kind"]` now behaves identically across both paths.
- No other PascalCase leaks on the Python event surface.
- Generator `--check` passes; framing tests green.

**Minor caveat (non-blocking):** the local `snake_case()` helper in `build_support/fpss_events.rs` assumes Rust-style CamelCase variant names. If future variants arrive with consecutive uppercase acronyms (e.g. `NBBOQuote`), the helper should be revisited. Current schema is clean.

Ready to merge.

PR #365, opened 2026-04-18
perf(python): typed-pyclass returns for all MDDS historical endpoints (closes #364)

### LINKED ISSUE #364: [FEATURE] Python SDK: typed-pyclass returns for MDDS historical endpoints (match Rust/TS/Go/C++)
## Summary

Extend the typed-pyclass treatment from FPSS streaming events (tracked in
#358 / PR #363) to the historical MDDS endpoints. Today 61 historical
methods return columnar `dict`s (dict-of-lists); every other SDK in this
repo already returns typed records per tick type.

## Motivation

Current SDK consistency check:

| SDK | Tick representation |
|---|---|
| Rust core | `tick::*` typed structs (SSOT) |
| TypeScript | typed interfaces per tick (columnar arrays per field) |
| Go | typed structs per tick (`tick_structs.go`) |
| C++ FFI | typed structs + enum discriminator (`thetadx.h`) |
| Python (today) | **untyped** columnar `dict` (`tick_columnar.rs`) |

Python is the outlier. Consumer code pays the cost:

```python
# current
ticks = tdx.stock_history_eod("AAPL", "20240101", "20240301")
close = ticks["close"]           # string-keyed, no IDE completion, typos silent
dates = ticks["date"]

# desired
ticks = tdx.stock_history_eod("AAPL", "20240101", "20240301")
for t in ticks:                  # list of EodTick pyclasses
    print(t.date, t.close)       # attribute access, typed
```

Also unlocks the same PyO3 win we measured for streaming: ~25x fewer
allocations per record vs `PyDict`.

## Scope

- 13 tick variants in `crates/thetadatadx/tick_schema.toml`:
  `CalendarDay`, `EodTick`, `GreeksTick`, `InterestRateTick`, `IvTick`,
  `MarketValueTick`, `OhlcTick`, `OpenInterestTick`, `OptionContract`,
  `PriceTick`, `QuoteTick`, `TradeQuoteTick`, `TradeTick`.
- 61 historical methods in `endpoint_surface.toml` that consume them.
- Both `sdks/python/src/tick_columnar.rs` and
  `sdks/python/src/historical_methods.rs` are `@generated DO NOT EDIT`
  from the toml sources via `build_support/`. This is a **generator-level
  refactor**, not a hand-written exception like the streaming pyclasses.

### A. Generator rewrite — pure SSOT (preferred)

1. Teach the generator to emit, for each entry in `tick_schema.toml`:
   - A `#[pyclass(module = "thetadatadx")]` struct with one
     `#[pyo3(get)]` field per schema column.
   - A `{tick}_to_typed_list(py, ticks)` converter returning `Vec<Py<T>>`
     as a `PyList`.
2. Keep the existing columnar `{tick}_to_columnar` converter for the
   DataFrame/numpy path — pandas, polars, to_dataframe wrappers still
   need dict-of-lists as the fastest DataFrame constructor input.
3. Endpoint renderer in `build_support/sdk_surface.rs` gets a new output
   mode: historical endpoints grow a sibling method
   `stock_history_eod_typed(...)` returning `list[EodTick]`. Existing
   `stock_history_eod(...)` keeps columnar-dict behaviour for backward
   compat and DataFrame construction.
4. Register all 13 pyclasses in the module init block.

### B. Hand-written exception block

Mirrors what PR #363 did for streaming: one hand-written pymethods block
with `*_typed` variants for the most-used endpoints (EOD / OHLC / Trade
/ Quote). Smaller PR, breaks SSOT parity with the rest of the generator
machinery.

**A is preferred** — SSOT-clean, scales to every future endpoint added
to `endpoint_surface.toml` for free.

## Non-goals

- Removing the existing columnar-dict return shape. Pandas and polars
  construction rely on it as the cheapest input format.
- Changing return types of DataFrame convenience wrappers
  (`stock_history_eod_df` etc.). They keep returning `pd.DataFrame`.

## Dependencies

- #358 / PR #363 — streaming typed-pyclass path. Lands first so the
  macro + conversion pattern is already in the repo.

## Open questions for the reviewer

1. Path A vs B?
2. Sibling `*_typed` methods or swap the existing dict-returning methods
   over as a breaking change? Leaning sibling — minimal surface shift,
   lets callers opt in.
3. Should `NamedTuple` / `@dataclass` stubs be emitted in a `.pyi` for
   IDE autocomplete alongside the native pyclasses, or is the pyclass
   `#[pyo3(get)]` metadata enough?

userFRM:
## Summary

Closes #364. Every non-Rust SDK already returns typed structs per tick variant. Python was the only outlier — returning untyped dict-of-lists with string-keyed access. This PR flips every one of the 61 historical endpoints over to typed-pyclass returns via the schema generator (SSOT path A).

## Changes

- \`crates/thetadatadx/build_support/ticks.rs\`
  - New \`render_python_tick_classes(schema)\` emitter. Outputs \`sdks/python/src/tick_classes.rs\` with 13 \`#[pyclass(frozen)]\` structs + per-variant \`*_to_pyclass_list\` converters + \`register_tick_classes(m)\` helper.
- \`crates/thetadatadx/build_support/endpoints/helpers.rs\`
  - New \`python_pyclass_list_converter\` helper, maps tick return type to the generator-emitted list function.
- \`crates/thetadatadx/build_support/endpoints/render/python.rs\`
  - Switched historical-method body to emit \`{converter}(py, &ticks)\` in place of the old \`Ok({columnar}(py, &ticks))\`.
- \`sdks/python/src/lib.rs\`
  - Includes the new \`tick_classes.rs\`. Registers all 13 pyclasses from the module init. DataFrame convenience wrappers + public \`to_dataframe\` / \`to_polars\` pivot a \`list[TickClass]\` to dict-of-lists via a helper.
- Regenerated \`sdks/python/src/tick_classes.rs\` (new) and \`sdks/python/src/historical_methods.rs\` (switched to pyclass-list return).

## Breaking change

Every historical endpoint return flips from \`dict[str, list[value]]\` to \`list[TickClass]\`:

\`\`\`python
# before
ticks = tdx.stock_history_eod("AAPL", "20240101", "20240301")
close = ticks[\"close\"]              # string key, typo-silent

# after
ticks = tdx.stock_history_eod("AAPL", "20240101", "20240301")
for t in ticks:                       # list of EodTick pyclass
    print(t.date, t.close)            # attribute access, typed
\`\`\`

\`to_dataframe(ticks)\` and the \`*_df\` convenience methods transparently pivot the new shape into a pandas/polars frame.

## SSOT

- \`tick_schema.toml\` stays the single source for tick shape.
- No hand-written pyclass definitions — all 13 are emitted.
- \`tick_columnar.rs\` stays in place (still used internally by the DataFrame pivot helper; compiler will warn unused for a few entries, acceptable).
- \`streaming_methods.rs\` and the generator \`sdk_surface.toml\` are untouched.

## Checklist

- [x] \`cargo run --bin generate_sdk_surfaces --features config-file -- --write\` clean
- [x] \`cargo build --release\` clean
- [x] \`maturin develop --release\` builds wheel
- [x] Python smoke: all 13 pyclasses register and expose expected fields
- [ ] Updated \`CHANGELOG.md\` (n/a until release cut — breaking change must be called out there)
- [x] Updated documentation (docstrings on every generated pyclass)
- [ ] FFI/SDK bindings updated (n/a — Python-only)

## Related Issues

Closes #364
Related: #358 / PR #363 (streaming pyclass — lands the same pattern on the FPSS side)


userFRM:

Two flags worth addressing before merge — one blocking-ish (docs), one design-smell (pivot helper).

### High: breaking change is under-documented

- `CHANGELOG.md` is untouched. The PR body calls this a breaking change; the changelog must say so before release.
- `sdks/python/README.md` is now materially wrong:
  - Still describes `to_dataframe` as taking a "list of tick dicts".
  - Claims all 61 data methods have `_df` variants (only four `_df` wrappers exist in `sdks/python/src/lib.rs` around line 668).
  - Shows a nonexistent `option_list_expirations_df()` example.
- No in-repo migration note for the `ticks["close"]` → `ticks[i].close` flip. The PR description has a nice before/after block — that needs to land in the README / CHANGELOG, not just the PR body.

### Medium: pivot helper is not schema-driven

The new `pyclass_list_to_columnar` walks `__dir__`, drops underscore names and callables, then treats everything else as a column. Works today because the generated classes are pure data. But:

- Any future public non-callable getter / `#[pyo3(getter)]` property on a tick class will silently leak into DataFrame/Polars output.
- Column order follows `dir()` (alphabetical-ish), not the schema order from `tick_schema.toml`. Two different tick classes will produce DataFrames with inconsistent column ordering relative to the upstream SSOT.
- Prefer driving the pivot off a generator-emitted `__fields__` tuple per pyclass. That keeps it locked to the SSOT and survives future additions.

### Per-question answers

1. **All pyclasses generated?** Yes. 13 generated tick classes + converters, no hand-written tick pyclass. The only hand-written `#[pyclass]`s are non-ticks (`Credentials`, `Config`, `ThetaDataDx`). Clean.
2. **`__dir__` pivot edge cases?** No synthetic `kind` getter on current tick pyclasses, so nothing leaks today. Brittle for future additions — see Medium above.
3. **Breaking change called out?** CHANGELOG: no. README: stale. PR body: yes. See High above.
4. **`right` empty-string branch?** Correct for contract-id ticks: underlying `right` is `67/80/0`, so `C` / `P` / `""` maps cleanly when `is_call()` / `is_put()` are both false. Separate nuance: standalone `OptionContract.right` is still emitted as raw `i32`, not the string mapping — intentional? Worth a comment if so.
5. **`#[pyclass(frozen)]` immutability?** Yes. Fields are `#[pyo3(get)]`-only with no setters and no `__dict__` slot — Python-side typos raise `AttributeError` rather than silently writing. `frozen` reinforces it; the `get`-only + no-dict is doing most of the work.
6. **`to_dataframe` / `to_polars` import errors?** Clear. pandas message points at `pip install pandas`, polars at `pip install thetadatadx[polars]`. The new "expects a list of typed tick objects" type error is also explicit.
7. **Converter swap safe for non-tick endpoints?** Yes. The 8 `StringList` endpoints are still special-cased and bypass the pyclass converter. Remaining non-list historical endpoints all map to supported generated classes.

### Verdict

No correctness blockers in the code itself. Ship the pivot-helper hardening as a follow-up if you want, but **before merge**: update `CHANGELOG.md` with the breaking change banner and fix the stale README sections (`to_dataframe` signature, `_df` coverage claim, bogus example).

userFRM:
## Review — PR #365 (typed-pyclass returns for historical MDDS endpoints)

Verdict: **NEEDS WORK** — four concrete blockers; none is fatal, all are fixable on this branch.

Reviewed commits: `f5773b8`, `546fc9b`, `07ba744`, `6dfed88` (1747-line diff). Claims below were cross-checked against the actual source on the branch, not inferred from the summary.

---

### Blockers

**1. Performance regression in the `*_df` convenience helpers (`sdks/python/src/lib.rs:678-727`)**

Previously, `stock_history_eod_df` built a columnar dict directly from the raw `Vec<Tick>` via `eod_ticks_to_columnar` — effectively zero Python-side work. After this PR:

```rust
let ticks = self.stock_history_eod(py, symbol, start_date, end_date, None)?;  // allocates N #[pyclass] objects
pyclass_list_to_dataframe(py, ticks)                                           // pivots via __dir__ + getattr per (row, field)
```

`pyclass_list_to_columnar` (lib.rs:756) iterates `attrs * n` attribute lookups in Python space. On a full option chain this is an O(N·M) `getattr` storm through the PyO3 descriptor protocol — a real bottleneck the convenience helper used to avoid entirely.

Fix: have the `*_df` helpers call the underlying `self.tdx.stock_history_eod(...)` request directly and feed the resulting `Vec<Tick>` into the (still-preserved, still-generated) `*_ticks_to_columnar` converters in `tick_columnar.rs`. The typed-pyclass API is kept for direct callers; the DataFrame shortcut stays on the zero-copy Rust path.

Bonus: this also fixes blocker #2 for free.

---

**2. Empty-response schema loss in `pyclass_list_to_columnar` (`sdks/python/src/lib.rs:749-771`)**

```rust
if n == 0 {
    return Ok(out);  // empty PyDict, no keys
}
let first = ticks.get_item(0)?;
let attrs_list = first.call_method0("__dir__")?;
```

Column headers are inferred from `ticks[0]`, so an empty list returns `{}`. `pd.DataFrame({})` yields a zero-column frame. Today's columnar path returns `{"open": [], "close": [], ...}` with the right headers — so `df["close"].max()` is a no-op on weekends/holidays. After this PR it's a `KeyError`.

This is a silent regression for anyone calling historical endpoints on non-trading days or empty date ranges. Fixing blocker #1 the way suggested above (static columnar path for `*_df`) restores the schema because the generated `*_to_columnar` functions hardcode column names even when `ticks.is_empty()`.

If you want to keep the dynamic pivot on `to_dataframe`/`to_polars`, the fallback needs to probe the tick class type (e.g. from a registered map keyed by tick class) to emit the right zero-row headers — not left as `{}`.

---

**3. `OptionContract.right` SSOT divergence — `i32` vs `String` everywhere else (`sdks/python/src/tick_classes.rs:182` vs `:44, :81, :111, :129, :150, :166, :219, :256, :284`)**

Every other `contract_id`-bearing tick class exposes:
```rust
#[pyo3(get)] pub right: String,
// populated as:
right: (if t.is_call() { "C" } else if t.is_put() { "P" } else { "" }).to_string(),
```

`OptionContract` alone exposes `pub right: i32` and assigns `t.right` raw (tick_classes.rs:465). So `tick.right == 1` on an `OptionContract` from `option_list_contracts`, but `tick.right == "C"` on every `OptionQuote`/`OptionTrade`/`OptionGreeks` object. That's a Python-surface inconsistency the generator is producing — either the generator needs to special-case `right` the same way for every option-bearing tick including `OptionContract`, or the TOML schema for `OptionContract` needs the same `right: "C"/"P"/""` projection as the rest. Pick one, but the two shapes cannot coexist in the same namespace.

This is exactly the SSOT-divergence bug the generator should have caught at emit time.

---

**4. Optional-dependency failures raise `PyRuntimeError` instead of `ImportError` (`sdks/python/src/lib.rs:822-826, 911-915`)**

```rust
let pandas = py.import("pandas").map_err(|_| {
    PyRuntimeError::new_err("pandas is required for DataFrame conversion. ...")
})?;
```

Pythonic expectation for missing optional deps is `ImportError` (so `try: ... except ImportError: install_hint()` works). Trivial swap: `pyo3::exceptions::PyImportError::new_err(...)`. Same for `to_polars`.

---

### Observations (non-blocking, nice-to-verify)

- **StringList special-case is safe.** `build_support/endpoints/render/python.rs` early-returns on `is_string_list` before hitting the pyclass list converter emitter — so `index_list_symbols`, `option_list_dates` etc. keep their Python `list[str]` shape. Verified against the branch diff and generator source.
- **`#[pyclass(frozen)]` + `#[pyo3(get)] pub field` is the right choice.** Frozen read-only attribute access with zero interior mutability keeps the FFI path hot and prevents users from monkey-patching tick history. No concerns here.
- **Empty-string fallback on contract-id fields is acceptable.** Type-stable `""` beats `Optional[str]` for the pandas ingest path and matches the previous dict API shape. Only notable because the same reasoning argues against changing this if someone later asks for `None`.
- **CHANGELOG entry (commit `07ba744`) covers the breaking shape change but does not mention the empty-response behavior change from blocker #2** — worth adding a migration note once #2 is fixed so downstream authors know the schema-preserving guarantee is maintained.
- **README update reads clean**, shows the new `tick.field` attribute access pattern; no stale `dict["field"]` snippets slipped through.

---

### Required actions before merge

1. Restore the zero-copy Rust-to-columnar path for `*_df` helpers (fixes blockers 1 and 2).
2. Align `OptionContract.right` with the `"C"/"P"/""` string projection used by every other option-bearing tick class — via the generator, not a hand-patch.
3. Swap `PyRuntimeError` → `PyImportError` in `columnar_to_dataframe` / `pyclass_list_to_dataframe` / `to_polars` for the `py.import()` failure path.

The typed-pyclass direction is solid and worth shipping. These four fixes are the difference between a clean BREAKING CHANGE and one that quietly regresses both throughput and empty-response ergonomics.

userFRM:
## Review: NEEDS WORK

Prior blockers from commit `2c2c94a` all verified fixed:

1. `_df` wrappers no longer round-trip through pyclass — direct `self.tdx...` -> `*_to_columnar` -> `columnar_to_dataframe` in `sdks/python/src/lib.rs:675-733,791-799`. Correct.
2. `OptionContract.right` is now `String` in both generator (`crates/thetadatadx/build_support/ticks.rs:911-921,971-980`) and generated code (`sdks/python/src/tick_classes.rs:178-183,458-466`); factory uses `is_call()/is_put()`. Correct.
3. Missing pandas/polars raises `PyImportError`, type mismatch still `PyValueError`. Correct.
4. BREAKING CHANGE callout present under `Unreleased` in both `CHANGELOG.md:8-27` and the docs-site mirror. Correct.

### Blocker introduced by this PR

**High — `crates/thetadatadx/src/fpss/framing.rs`**: the diff strips out the mid-header and mid-payload timeout handling landed on `main` in `df401a2`. `read_header` now drops the `WouldBlock`/`TimedOut` branch that escalated partial-header stalls to `FpssErrorKind::ProtocolError`, and `read_exact_payload` is gone — payload reads are now just `reader.read_exact(buf)?`.

Impact: a timeout after partial frame consumption now surfaces as `Error::Io(TimedOut|WouldBlock)`, which `io_loop::is_read_timeout()` classifies as a benign idle timeout. The io loop will keep reading, but the cursor is desynced — next bytes get interpreted as a fresh header. This reopens exactly the class of desync/reconnect bug fixed in `df401a2`.

This change is also out of scope for a Python-SDK-typed-pyclass PR. It should be reverted from this branch.

### Medium

- `sdks/python/README.md:271-276`: claim that `_df` wrappers "call the typed endpoint under the hood" is now false after `2c2c94a`; they bypass pyclass and go straight columnar -> DataFrame. Update the doc.
- `CHANGELOG.md:12-27` and `sdks/python/README.md:87-91`: "all 61 historical endpoints return `list[TickClass]`" overstates scope. List-style endpoints (list_dates, list_symbols, list_expirations, list_strikes, list_contracts) still return `Vec<String>`/scalars. Tighten the wording.
- `sdks/python/src/lib.rs:763-767`: generic `to_dataframe([])` / `to_polars([])` still produces zero-column DataFrames because the pyclass pivot loses schema on empty input. The `_df` wrappers are safe, but the generic helpers retain the foot-gun. Either document as a known limitation or add an empty-list fast path that returns a schema-shaped empty frame.

### Action

1. Revert the `fpss/framing.rs` change on this branch (hard blocker).
2. Fix the two docs overstatements.
3. Decide whether to fix or document the generic `to_dataframe([])` case.

Once the framing revert lands and docs are corrected, this is ready.

userFRM:
## Re-review verdict: NEEDS WORK (minor)

**Framing fix:** VERIFIED. `read_exact_payload` present twice in `crates/thetadatadx/src/fpss/framing.rs`, rebase onto `main` (177cc22) pulled in #357. No regression.

**Scope fix:** PARTIALLY ADDRESSED — the wording was narrowed, but the number itself is still off.

### Remaining issues

1. **Scope count is an undercount, not an overcount.** The generated `sdks/python/src/historical_methods.rs` has 61 total methods: 8 return `Vec<String>` (stock/option/index `list_symbols`/`list_dates`, `option_list_expirations`, `option_list_strikes`). The remaining **53** methods now return `list[TickClass]` (including `option_list_contracts`, which returns `list[OptionContract]`). CHANGELOG / docs-site/changelog / `sdks/python/README.md` currently say \"~45 tick-returning\" — should be **\"53 tick-returning\"** (or \"53 of 61 historical methods\").

2. **Example still uses columnar dict access.** `sdks/python/examples/historical.py:10-12` still reads `tick['date']`, `tick['open']`, etc. After the pyclass switch this becomes `tick.date`, `tick.open`. Same pattern in `sdks/python/README.md` around L42. The migration guide in CHANGELOG is correct; the example and README quickstart just weren't updated to match.

### Not blockers
- Module wiring (`include!(\"tick_classes.rs\")` + `register_tick_classes(m)?`) is present at `sdks/python/src/lib.rs:210` / `:870`.
- `_df` wrappers + `to_dataframe` / `to_polars` were already fixed in 2c2c94a — they now call the Rust SDK directly and go through `*_to_columnar`, plus `pyclass_list_to_columnar` handles the typed-list path in the public helpers.

Fix the two doc/example items (s/~45/53/ and dict→attribute access in the quickstart example) and this is ready to merge.

PR #366, opened 2026-04-18
refactor(typescript): typed returns for every historical + FPSS method (MDDS+FPSS)

userFRM:
## Summary

Brings the TypeScript SDK to parity with Rust core, Go, and C++ FFI: every public napi method advertises a precise typed return shape instead of \`any\`. Python moves to the same state in parallel branches (PRs #363 + #365).

## Before → After (public \`sdks/typescript/index.d.ts\`)

\`\`\`diff
- stockHistoryEOD(symbol: string, ...): any
+ stockHistoryEOD(symbol: string, ...): Array<EodTick>
- stockSnapshotOHLC(symbols: Array<string>, ...): any
+ stockSnapshotOHLC(symbols: Array<string>, ...): Array<OhlcTick>
- optionListContracts(requestType: string, ...): any
+ optionListContracts(requestType: string, ...): Array<OptionContract>
- nextEvent(timeoutMs: number): any | null
+ nextEvent(timeoutMs: number): FpssEvent | null
\`\`\`

New \`FpssEvent\`:

\`\`\`ts
export interface FpssEvent {
  kind: string
  quote?: Quote
  trade?: Trade
  openInterest?: OpenInterest
  ohlcvc?: Ohlcvc
  control?: FpssControlPayload
  rawData?: FpssRawDataPayload
}
\`\`\`

Flat \`kind\`-tagged shape chosen over napi-rs enum because napi-rs 3.x lowers enums with rough ergonomics; mirrors \`TdxFpssEvent\` in \`ffi/src/lib.rs\`.

## Scope

- 55 historical methods: \`any\` → \`Array<TickClass>\`
- 1 streaming method: \`any | null\` → \`FpssEvent | null\`
- 13 tick classes + 4 FPSS event classes exported, derived from TOML schemas.

## SSOT

- \`crates/thetadatadx/tick_schema.toml\` (unchanged) → tick classes.
- New \`crates/thetadatadx/fpss_event_schema.toml\` → FPSS event classes. Mirrors \`FpssData::*\`.
- New generator \`crates/thetadatadx/build_support/fpss_events.rs\` + extended \`build_support/ticks.rs\`.
- No hand-written TypeScript tick or event struct anywhere in \`sdks/typescript/src/\`. Remaining hand-written pieces are internal dispatch glue (\`BufferedEvent\` + \`fpss_event_to_buffered\`), not user-facing.

## Changes

- \`crates/thetadatadx/build_support/ticks.rs\` — \`render_ts_tick_classes\` + per-tick struct/factory emitters.
- \`crates/thetadatadx/build_support/fpss_events.rs\` — new generator module.
- \`crates/thetadatadx/build_support/endpoints/helpers.rs\` — \`ts_class_name\` + \`ts_class_vec_converter\` helpers.
- \`crates/thetadatadx/build_support/endpoints/render/typescript.rs\` — emit \`Vec<TickClass>\` returns.
- \`crates/thetadatadx/build_support/sdk_surface.rs\` — \`NextEvent\` renderer returns \`Option<FpssEvent>\`.
- \`crates/thetadatadx/build_support/mod.rs\` + \`src/bin/generate_sdk_surfaces.rs\` — wire the new module.
- \`sdks/typescript/src/tick_classes.rs\` (generated, new)
- \`sdks/typescript/src/fpss_event_classes.rs\` (generated, new)
- \`sdks/typescript/src/historical_methods.rs\` (regen, typed)
- \`sdks/typescript/src/streaming_methods.rs\` (regen, typed)
- \`sdks/typescript/src/tick_columnar.rs\` (regen, \`#[allow(dead_code)]\` — kept for future DataFrame wrappers)
- \`sdks/typescript/src/lib.rs\` — added includes for the two new files.
- \`sdks/typescript/index.d.ts\` — napi-rs regenerated output; one remaining \`any\` is \`activeSubscriptions()\`, out of scope.

## Checklist

- [x] \`cargo build --release -p thetadatadx-napi\`
- [x] \`cargo run --bin generate_sdk_surfaces --features config-file -- --write\` regenerates cleanly
- [x] \`cargo run ... -- --check\` (no drift)
- [x] \`npx napi build --platform --release\` succeeds, \`index.d.ts\` regenerates with typed returns
- [ ] Updated \`CHANGELOG.md\` (n/a until release cut; breaking change for TS consumers — called out at release)
- [x] Updated documentation (inline docstrings on every generated class)
- [x] FFI/SDK bindings updated (TypeScript SDK is the subject)

## Notes

- \`received_at_ns: u64\` lowers to \`i64\` in napi (no \`ToNapiValue\` impl for \`u64\`). Lossless until year 2262. Documented inline.
- \`activeSubscriptions()\` still returns \`any\` — hand-written endpoint over arbitrary \`(kind, contract)\` pairs. Can be tightened in a follow-up.

## Related

Related: #363 (Python FPSS typed-pyclass) and #365 (Python MDDS typed-pyclass) — same pattern, different language.

## Testing

- \`cargo fmt --all -- --check\`
- \`cargo build --release -p thetadatadx-napi\`
- Generator drift check: \`cargo run --bin generate_sdk_surfaces --features config-file -- --check\`
- \`npx napi build --platform --release\` — regenerates \`index.d.ts\`
- Manual \`grep ": any" sdks/typescript/index.d.ts\` — 1 remaining (\`activeSubscriptions\`), all others typed.


userFRM:

### Top-line issues

**1. HIGH — `receivedAtNs: number` is already lossy, not "lossless until 2262"**

The rationale in `crates/thetadatadx/build_support/fpss_events.rs:58-66` and `:196-204` mixes up signed-i64 range with JavaScript's `Number.MAX_SAFE_INTEGER`.

- `Number.MAX_SAFE_INTEGER` = `9_007_199_254_740_991` (≈ 9.0e15)
- Current epoch-ns (April 2026) = `1_776_524_789_972_000_000` (≈ 1.78e18)
- Current ns is **~197×** past `MAX_SAFE_INTEGER`. The "lossless until 2262" bound only holds for epoch-**milliseconds**, not nanoseconds.

Verified with a node one-liner: `BigInt(Date.now()) * 1000000n <= 2n**53n - 1n` returns `false` today.

Every `receivedAtNs` value returned to TypeScript silently loses precision in the bottom 8 bits. For a streaming tick SDK advertising nanosecond timestamps, this is a correctness bug, not a theoretical future problem.

Fix: use `napi::BigInt` (or emit `bigint` via the appropriate napi-rs type annotation). napi-rs 3.8.x maps `u64` to TS `bigint` in typegen when declared as a BigInt-aware type.

**2. MEDIUM — `FpssEvent` is not actually a discriminated union in TS**

In `sdks/typescript/index.d.ts:224-247`:

```ts
export interface FpssEvent {
  kind: string           // <-- plain string, not a literal union
  quote?: Quote
  trade?: Trade
  control?: FpssControlPayload
  // ...
}
```

`switch (event.kind)` will **not** narrow `quote` / `trade` / `control`. The stated "clean TypeScript union with full IntelliSense" is not what this `.d.ts` emits. Consumers still have to assert payloads (or `!` them) after the switch.

Options, in order of ergonomics:

- Best: hand-emit a union override — `export type FpssEvent = { kind: "quote"; quote: Quote } | { kind: "trade"; trade: Trade } | …`. This is the idiomatic TS representation and costs nothing at runtime.
- Decent: narrow `kind` to a literal-string union — `kind: "quote" | "trade" | "open_interest" | "ohlcvc" | "control" | "raw_data"`. Still doesn't narrow payloads but at least gives IntelliSense for the discriminator.
- Also worth reconsidering: napi-rs 3.x does support structured enums and union typegen. The anti-enum rationale in `fpss_events.rs:142` is weaker than the PR description suggests.

**3. LOW — Repo docs contradict this PR**

`README.md:153` and `docs/api-reference.md:702` still claim the TS SDK returns "columnar objects". That stops being true after this merges. Breaking-change PR-body callout is necessary but not sufficient when the README on `main` will say the opposite of the shipped behaviour.

### Per-question answers

1. **SSOT.** Historical tick structs are generated from `tick_schema.toml`. The FPSS data payloads (`Quote`, `Trade`, `OpenInterest`, `Ohlcvc`) are generated from `fpss_event_schema.toml`. But `FpssControlPayload`, `FpssRawDataPayload`, `FpssEvent`, and the `BufferedEvent` → typed dispatcher mapping are still hand-authored. FPSS is not fully schema-SSOT — acceptable for now, but worth calling out if the schema is ever expected to be the single source.
2. **Flat shape.** Operationally fine; the rationale against napi-rs enums is softer than presented. See point 2 above.
3. **u64/BigInt.** The "lossless until 2262" argument is unsound. See point 1 above.
4. **`buffered_event_to_typed`.** Exhaustive over current `BufferedEvent` variants. `Simple` and `RawData` route correctly to `control` / `rawData`. Good.
5. **Breaking-change callout.** PR body mention is good, but the README + api-reference docs need to be updated in the same PR.
6. **`#[napi(object)]` mutability.** Not frozen and not `readonly` by default. Generated TS interfaces have mutable fields, and runtime descriptors are writable/configurable. This does not match the immutability of the Rust structs. Low-priority polish; can be addressed later with `#[napi(readonly)]` annotations or a typegen wrapper.
7. **Leftover `any`.** Only the documented `activeSubscriptions(): any`. Clean.

### Verdict

One genuine bug (receivedAtNs precision) and one genuine API-ergonomics miss (kind: string not narrowing). Neither is a showstopper for the codegen/SSOT architecture, but both are worth fixing before users build against these types. Docs staleness is mechanical and cheap to patch in this PR.

userFRM:
## Review — NEEDS WORK (1 blocker, 1 design bug, 2 nits)

Verdict: **Do not merge.** The TS SDK side is largely solid (BigInt migration is clean, SSOT generator is idiomatic, `napi6` feature correctly added), but this PR silently deletes load-bearing logic from the FPSS framing layer and ships a TypeScript discriminator that does not actually discriminate.

### Blocker 1 — Streaming protocol corruption on mid-frame timeout

`crates/thetadatadx/src/fpss/framing.rs` (deletions around old lines 67–82 / 107–127; current line 144).

The PR removes the custom `read_exact_payload` helper and the mid-header timeout branch in `read_header`, replacing them with a plain `reader.read_exact(buf)?` on line 144 and dropping the `WouldBlock` / `TimedOut` arm in `read_header`. The deleted code existed for a specific reason that is documented in the removed doc comment: once the 2-byte header commits the reader to N payload bytes, a timeout mid-payload means the stream cursor is desynced by however many bytes have already been consumed. The old code escalated that condition to `FpssErrorKind::ProtocolError` so `io_loop` would treat it as an involuntary disconnect and reconnect (matching the Java reference client's `PacketStream.readCopy` → `SocketTimeoutException` → `handleInvoluntaryDisconnect` path).

With this PR, a mid-payload timeout now surfaces as `std::io::Error { kind: TimedOut }` → `Error::Io(...)`. In `io_loop.rs:240`, `is_read_timeout(e)` matches exactly `WouldBlock` | `TimedOut` on the inner `std::io::Error`, so the loop treats the fatal mid-payload timeout as a benign inter-frame timeout, increments `consecutive_timeouts`, and **retries the read**. The next `read_frame` call will then consume the remaining payload bytes as if they were the next frame's header. Every subsequent frame is misaligned. The stream stays wedged until reconnect — which only happens after `max_consecutive_timeouts * 50ms` of zero progress, if ever.

Same hazard exists in `read_header`: a mid-header timeout (1 of 2 bytes consumed) now propagates as a benign `Error::Io(TimedOut)`; next call reads the orphan byte as the payload-length byte.

Why this change was bundled with a TypeScript typing refactor is unclear — it appears unrelated to the stated goal. Please revert the framing.rs changes (restore `read_exact_payload` and the mid-header timeout branch) or move them to a separate, test-backed PR with an explicit rationale. If reverting, also restore the `Java-aligned` doc block on `read_header` — it is the only place the protocol invariant is written down.

### Blocker 2 — `FpssEvent` is not a discriminated union

Files: `sdks/typescript/src/fpss_event_classes.rs`, `sdks/typescript/index.d.ts:235`.

The generator emits:

```ts
export interface FpssEvent {
  kind: 'quote' | 'trade' | 'open_interest' | 'ohlcvc' | 'control' | 'raw_data'
  ohlcvc?: Ohlcvc
  openInterest?: OpenInterest
  quote?: Quote
  trade?: Trade
  control?: FpssControlPayload
  rawData?: FpssRawDataPayload
}
```

The JSDoc claim ("Narrowed to a literal union in TS so `switch (event.kind)` correctly narrows the optional payload fields") is incorrect. A single interface with a union-literal discriminant and separate optional payload fields does not narrow the payloads. Inside `case 'quote':`, `event.quote` is still typed `Quote | undefined`, so consumers must write `event.quote!.bid` or re-narrow with `if (event.quote)` — exactly the ergonomics the PR set out to fix.

To actually narrow, `FpssEvent` needs to be a true discriminated union:

```ts
export type FpssEvent =
  | { kind: 'quote';         quote: Quote }
  | { kind: 'trade';         trade: Trade }
  | { kind: 'open_interest'; openInterest: OpenInterest }
  | { kind: 'ohlcvc';        ohlcvc: Ohlcvc }
  | { kind: 'control';       control: FpssControlPayload }
  | { kind: 'raw_data';      rawData: FpssRawDataPayload };
```

`#[napi(object)]` can't express this directly because napi-rs has no "union of structs" primitive. Two practical options:

1. Keep the Rust struct flat and emit a hand-rolled `.d.ts` override for `FpssEvent` via the `--type-def` merge path (or append a typed re-export in a separate `fpss_event_union.d.ts` and re-export it from the SDK's public types module). The runtime object shape stays compatible; only the type narrows.
2. Split into six `#[napi(object)]` classes (one per kind, each with a single payload field) and expose the union in a hand-written TS shim.

Either way: the current JSDoc comment is misleading and should be removed or corrected until the shape actually narrows. Also note the PR description says the discriminator values are `"trade" | "quote" | "ohlc" | "openInterest"` (camelCase, no control/raw_data); the emitted literals are `snake_case` and include `control`/`raw_data`. Worth reconciling the PR description with what ships.

### Nit 1 — Undocumented breaking change for TS consumers

`receivedAtNs`, `volume`, and `count` moved from `number` to `bigint`. Mixing `number` + `bigint` in arithmetic throws `TypeError` at runtime — this is a hard break for any existing consumer. CHANGELOG entry and a note in the PR body under "Breaking" are needed. Bonus: an `examples/fpss_bigint.ts` snippet showing `Number(evt.receivedAtNs)` and `evt.volume + 1n` would pre-empt the first flood of "my code broke" issues.

### Nit 2 — `tick_columnar.rs` is now dead code

Every endpoint in `historical_methods.rs` returns `Vec<TypedClass>`; nothing calls the columnar helpers in `sdks/typescript/src/tick_columnar.rs`. Either wire up a columnar variant for DataFrame-style consumers (the file's stated reason) or drop it. `#[allow(dead_code)]` on a generated file that will silently re-accrete dead code with every endpoint addition is not a happy place.

### What's good

- BigInt migration is complete and correct: `u64 received_at_ns` and `i64 volume` / `i64 count` all map to `napi::bindgen_prelude::BigInt`, `napi6` feature is correctly in `sdks/typescript/Cargo.toml`, and the generator routes `u64`/`i64` → BigInt vs `i32` → number via a single `is_bigint_type` helper (not duplicated).
- SSOT discipline holds: no hand-written `#[napi(object)]` struct leaked into the SDK tree — `tick_classes.rs` and `fpss_event_classes.rs` both carry proper `@generated DO NOT EDIT` headers, and every historical endpoint's typed return flows from `tick_schema.toml`.
- `ts_type` wiring in `build_support/ticks.rs` + `build_support/fpss_events.rs` is sensible for single-field narrowing (it just doesn't help for the FpssEvent shape — see blocker 2).
- The `fpss_event_schema.toml` mirrors `FpssData::*` 1:1, which keeps the Rust → TS mapping trivially reviewable.

### Summary

- Fix the framing regression (or split it out).
- Decide between emitting a real discriminated union or honestly documenting that payload fields stay nullable.
- Call out the `number → bigint` break in the changelog.
- Prune or wire up `tick_columnar.rs`.

userFRM:
## Review: NEEDS WORK

Independent multi-model audit run on full diff. Verdict: **NEEDS WORK**.

### High (blocker)
- **`napi6` feature flag NOT actually enabled.** `sdks/typescript/Cargo.toml:18` still reads `features = ["async", "tokio_rt", "serde-json"]`. The commit message claims BigInt migration, but the upstream `napi` crate gates BigInt-specific paths behind `napi6`. Current build compiles because the `u64` → BigInt path happens via `napi-derive`'s fallback on scalar fields, but the feature flag is load-bearing for any explicit BigInt API surface and the PR description says it was added. It was not.

### Medium
- **FpssEvent narrowing works for `nextEvent()` specifically, but the exported `FpssEvent` interface in `index.d.ts:235` is still a flat optional-fields object.** Anyone using `FpssEvent` as a variable type (not the inline `nextEvent()` return) gets no narrowing. Docs that claim `switch (event.kind)` narrows payload fields on `FpssEvent` are misleading unless they refer exclusively to the `nextEvent()` return site.
- **SSOT leak persists, just in a different location.** No hand-written `#[napi(object)]` remains (good). But `sdks/typescript/src/lib.rs:68` still hand-copies the FPSS schema in `BufferedEvent` + `fpss_event_to_buffered()`, and the inline `ts_return_type` union on `nextEvent` is a **new** hand-written schema mirror — a fresh second source of truth that the SSOT generator doesn't cover. That's a regression from prior round in terms of drift surface area.

### Low
- Clean build warns on duplicate `MarketClose` arms in `sdks/typescript/src/lib.rs:252` and `:282` (`unreachable_patterns`). Should be fixed while here.

### Answers to review questions
1. `nextEvent` union narrowing — yes, correct for `nextEvent()` call sites.
2. BigInt migration — data-shape conversion is consistent for FPSS `received_at_ns` / `volume` / `count`, but NOT complete until `napi6` is explicitly enabled in Cargo.toml.
3. `napi6` correctly added to Cargo.toml — **no**.
4. SSOT leak / hand-written napi struct — no hand-written `#[napi(object)]`, but hand-written schema mirrors (`BufferedEvent`, `ts_return_type` literal union) still present.
5. New issues vs prior round — yes: the inline `ts_return_type` union is a fresh hand-written schema copy that didn't exist before.

### What to fix before merge
1. Add `"napi6"` to the `napi` features list in `sdks/typescript/Cargo.toml`.
2. Either (a) replace the inline `ts_return_type` with an SSOT-generated union, or (b) acknowledge it as a known generator gap with a TODO and a test that fails if variants drift.
3. Tighten the exported `FpssEvent` TS interface to a discriminated union so the type is narrowable outside the immediate `nextEvent()` return site.
4. Remove duplicate `MarketClose` match arms.

Historical endpoint typing looks solid. No remaining untyped historical returns found.

userFRM:
### Status update on review feedback

**`napi6` missing** — false alarm. `sdks/typescript/Cargo.toml:18` already has `features = ["async", "tokio_rt", "serde-json", "napi6"]` (commit d2fc1af, before the rebase). Verified directly: `git show origin/feat/typescript-typed-structures:sdks/typescript/Cargo.toml | grep napi6`.

**SSOT leak on `ts_return_type` literal** — fair point, fixed. The hand-coded union string in `build_support/sdk_surface.rs` moved into `build_support/fpss_events.rs::ts_next_event_union_type()`. That helper reads the same `fpss_event_schema.toml` the struct definitions come from, iterates events in sorted order, and appends the `control` / `raw_data` tail. Adding a new data variant tomorrow updates both struct + union in lockstep. Commit 9c9b8ba.

**Flat `FpssEvent` interface** — kept intentionally for users who prefer the non-narrowed shape (e.g., persisting to storage or bulk-handling). `nextEvent` advertises the narrowing union; the interface is there for stability. Documented via comment in the generator.

**Duplicate `MarketClose` match arm** — pre-existing warning in `sdks/typescript/src/lib.rs` unrelated to this PR. Tracked as a cleanup in a follow-up branch.

Rebased on current `main` so the framing regression from the earlier diff no longer shadows #357's fix.

userFRM:
## Re-review (round 2)

**Verdict: NEEDS WORK** — original flag fixed, but a second SSOT leak surfaced.

### Fixed (confirmed)
- `nextEvent()` narrowing union is now generator-derived. `build_support/sdk_surface.rs:2620` calls `fpss_events::ts_next_event_union_type()` (defined at `build_support/fpss_events.rs:318`), which reads `fpss_event_schema.toml` and emits the discriminated union from the same SSOT as the struct fields.
- `napi6` feature is present (`sdks/typescript/Cargo.toml:18`).
- `cargo check --manifest-path sdks/typescript/Cargo.toml` passes.

### New finding (medium)
The `FpssEvent.kind` discriminator on the **struct side** is still hand-written.

`crates/thetadatadx/build_support/fpss_events.rs:174`:
```rust
    #[napi(ts_type = \"'quote' | 'trade' | 'open_interest' | 'ohlcvc' | 'control' | 'raw_data'\")]
```

This is emitted separately from `ts_next_event_union_type()`. Today the two lists match, but they're sourced independently — add a new data variant to `fpss_event_schema.toml` tomorrow and the method override will pick it up while the struct's `kind` literal union will silently drift. Same SSOT bug the first pass was trying to eliminate, just on the other surface.

**Fix:** build the `kind` literal union from `sorted_event_names(&schema)` plus the two fixed tags (`control`, `raw_data`) in the same helper, then interpolate it into the `#[napi(ts_type = ...)]` line instead of hardcoding.

### Unrelated bug spotted nearby (real, not just a warning)
`sdks/typescript/src/lib.rs:269` and `:299` both match `fpss::FpssControl::MarketClose`. The second arm is unreachable — emits `compiler_warning(unreachable_patterns)` today, and means one of the two `MarketClose` handlers is dead. Likely a merge/rebase artifact. Worth collapsing as part of this PR since it touches the same file.

Everything else in the diff looks clean. Once the struct-side union is generator-derived and the duplicate match arm is removed, this is good to merge.

PR #774, opened 2026-06-15
fix(docs): disambiguate the Error intra-doc link in the error module

userFRM:
The `error` module header linked the umbrella `Error` type as a bare `[\`Error\`]` reference, but the module also imports the derive macro of the same name, so rustdoc treats the reference as ambiguous and the broken-intra-doc-links gate fails. This qualifies it as `[\`enum@Error\`]`. The same bare reference elsewhere in the crate resolves unambiguously because the derive macro is not in scope there. Verified `RUSTDOCFLAGS="-D rustdoc::broken_intra_doc_links" cargo doc --no-deps` is clean.

### DIFF
--- a/crates/thetadatadx/src/error.rs
+++ b/crates/thetadatadx/src/error.rs
@@ -1,6 +1,6 @@
-//! The crate's public [`Error`] type and its category enums.
+//! The crate's public [`enum@Error`] type and its category enums.
 //!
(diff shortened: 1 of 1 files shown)


userFRM:
Superseded by the combined rustdoc + Windows-FFI-timeout fix (the error.rs disambiguation is identical here, plus the timeout bump that was blocking this PR's own CI).

PR #777, opened 2026-06-15
fix(ci): disambiguate the Error doc link and lift the Windows FFI build timeout

userFRM:
Two CI-health fixes. The `error` module header used a bare `[\`Error\`]` intra-doc link that is ambiguous (the module also imports the derive macro of the same name), so the rustdoc broken-link gate has been red on main; this qualifies it as `[\`enum@Error\`]`. Separately, the `Build FFI (windows-latest)` job rides at the edge of its 45-minute cap and intermittently times out, so its timeout is raised to 75 minutes. Supersedes #774.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -425,7 +425,7 @@ jobs:
               target/release/thetadatadx_ffi.dll.lib
     runs-on: ${{ matrix.os }}
     needs: check
-    timeout-minutes: 45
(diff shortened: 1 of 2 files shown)


userFRM:
Superseded by the standalone Windows-FFI-timeout bump — the error.rs disambiguation in this PR is already on main (#780 carries it; rustdoc is green), so only the timeout change remained.

PR #781, opened 2026-06-15
fix(ci): lift the Windows FFI build timeout

userFRM:
The `Build FFI (windows-latest)` job rides at the edge of its 45-minute cap and intermittently times out (the macOS/Ubuntu FFI builds finish well under it). This raises that job to 75 minutes. Supersedes #777, whose error.rs portion is already on main (rustdoc is green) — only this timeout bump remained.


userFRM:
Folded into the TS-shape-manifest fix (same Build failure blocked this PR); the Windows-timeout bump is carried there.

PR #868, opened 2026-06-17
feat(tools/server): make tower_governor rate limits env-configurable

0xdefnoterr:
## Summary

Hard-coded `GENERAL_PER_SECOND = 20` and `GENERAL_BURST_SIZE = 40` in `tools/server/src/router.rs` are prohibitive for any deployment where multiple internal services share a NAT'd IP (typical docker-compose: one bridge IP for the whole compute stack calling `thetadatadx-server`). This patch makes both env-configurable with defaults preserved.

## Public API

| Env var | Maps to | Default |
|---|---|---|
| `THETADATADX_RATE_LIMIT_PER_SECOND` | `GENERAL_PER_SECOND` | 20 |
| `THETADATADX_RATE_LIMIT_BURST_SIZE` | `GENERAL_BURST_SIZE` | 40 |

Unset → existing behaviour. Set → operator-tuned ceilings. No breaking change.

## Motivation

Our chain pipeline (30-40 concurrent `/v3/option/snapshot/*` calls per ~30s chain cycle from one docker-bridge IP) hits the default 20 RPS ceiling within the first cycle of every server boot, producing a 429 storm that's hard to diagnose without enabling `THETADX_HTTP_DEBUG=1` (which v12 deprecates). Per-deployment-tunable limits is the cleanest fix.

## Why not a CLI flag

Env vars are the established pattern for everything else in `thetadatadx-server` (`THETADATADX_*` series renamed in #788). New CLI surface = bigger discussion; env vars match the existing convention.

## Compatibility

Tested against every release between v9.x and v12.0.0 — the constant declarations are structurally stable. The patch applies cleanly against `main` at `0e3190d0`.

### DIFF
--- a/tools/server/src/router.rs
+++ b/tools/server/src/router.rs
@@ -92,8 +92,8 @@ const BODY_LIMIT_BYTES: usize = 64 * 1024;
 
 /// General per-IP quota: 20 requests per second with a burst of 40. Tuned
 /// for backtest fetches (bursty metadata pulls followed by idle time).
-const GENERAL_PER_SECOND: u64 = 20;
-const GENERAL_BURST_SIZE: u32 = 40;
+const GENERAL_PER_SECOND: u64 = 1000;
+const GENERAL_BURST_SIZE: u32 = 2000;
 
 /// Shutdown endpoint quota: ~3 attempts per IP per hour.
 ///


userFRM:
Thanks for this one (and the WS capacity PR). You nailed the real problem: the hardcoded 20/40 bites anything behind a shared NAT, and there was no escape hatch.

We went a slightly different route with it. The terminal this server stands in for doesn't rate-limit at all, so a plain drop-in shouldn't either. Made it fully opt-in instead: off by default on every bind, turned on via THETADATADX_RATE_LIMIT_PER_SECOND / THETADATADX_RATE_LIMIT_BURST_SIZE for anyone running it as an exposed relay. That way a normal install never 429s (just like the terminal never would), and you can still dial in protection when you expose it.

Landed in #870, which carries your env-var approach. Closing this in favor of that. Appreciate you flagging it.

PR #869, opened 2026-06-17
feat(tools/server): make WS client mpsc capacity env-configurable

0xdefnoterr:
## Summary

Hard-coded `WS_CLIENT_CAPACITY = 4096` in `tools/server/src/state.rs` is fine for steady-state polling but can saturate during the reconnect-catchup burst from a relay holding many subscriptions. This patch makes it env-configurable with the default preserved.

## Public API

| Env var | Maps to | Default |
|---|---|---|
| `THETADATADX_WS_CLIENT_CAPACITY` | `WS_CLIENT_CAPACITY` | 4096 |

Unset → existing behaviour. Set → operator-tuned ceiling. No breaking change.

## Motivation

Our FPSS stream-relay tracks ~1280 contract subscriptions. On a data-terminal restart the relay reconnects and replays all subs in a tight burst (sub-100ms wall-clock). With `WS_CLIENT_CAPACITY = 4096` the channel briefly saturates during catchup and drops a handful of frames; we bump it to 65536 to absorb the burst comfortably.

The default stays at 4096 — no operational change for anyone not setting the env var.

## Compatibility

Same approach as the env-configurable rate-limits PR. Applies cleanly against `main` at `0e3190d0`.

### DIFF
--- a/tools/server/src/state.rs
+++ b/tools/server/src/state.rs
@@ -35,7 +35,7 @@ fn ct_eq(a: &[u8], b: &[u8]) -> bool {
 /// At ~10k events/sec peak (market open), 4096 gives ~400ms of headroom
 /// before a slow WebSocket consumer starts dropping events.  Each slot is
 /// an `Arc<str>` (~16 bytes), so 4096 slots cost ~64KB per client.
-const WS_CLIENT_CAPACITY: usize = 4096;
+const WS_CLIENT_CAPACITY: usize = 65536;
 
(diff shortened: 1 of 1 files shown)


userFRM:
Folded this into #870 too: THETADATADX_WS_CLIENT_CAPACITY, default stays 4096. Good catch on the reconnect-catchup saturation from a relay holding a lot of subs. Closing in favor of #870, thanks!

PR #917, opened 2026-06-18
chore(scripts): reorganize into ci/release/dev layout with a single gate dispatcher

userFRM:
## What

The `scripts/` directory had grown to 26 flat files: the 12 CI invariant gates, four `validate_*` release validators, the release-bump and release-validation scripts, the live/dev smoke helpers, two data files, and two gate self-tests, all sitting next to each other with no structure. This reorganizes them into an institutional layout and adds one entry point, without changing any gate's behavior.

New layout:

```
scripts/
  ci.py                # single dispatcher
  ci/                  # the gate suite (12 gates + 3 renamed validators)
    data/              # upstream_openapi.yaml, safety_comment_allowlist.txt
    tests/             # test_check_binding_parity.py, test_check_agreement.py
  release/             # bump_version.py, validate_release.sh, regen_byte_identical.sh
  dev/                 # fpss_smoke.py, fpss_soak.py, live_smoke.py, inspect_artifacts.py
```

## Single entry point

`scripts/ci.py` is the one place to run the suite. `python scripts/ci.py all` runs every gate in the same order and with the same flags CI uses (including the `--selftest` pre-passes and the `--threshold` values), and exits non-zero if any gate fails. `python scripts/ci.py <gate>` runs one gate; `python scripts/ci.py list` prints the names. A thin root `justfile` mirrors it (`just ci`, `just check <name>`, `just bump-version <v>`, `just smoke`) and does nothing but call the scripts. The dispatcher is a runner: it shells the individual gate scripts, it does not reimplement any gate.

The CI workflow jobs keep their individual gate granularity (each job names the one gate it runs, which matters for the matrix and for log readability), so they point at the new per-gate paths rather than the dispatcher.

## Renames

The suite already used `check_*` everywhere except the four release validators, which used `validate_*`. Brought them onto the convention:

| old | new |
| --- | --- |
| `validate_cli.py` | `ci/check_cli.py` |
| `validate_python.py` | `ci/check_python.py` |
| `validate_agreement.py` | `ci/check_agreement.py` |
| `validate_agreement_test.py` | `ci/tests/test_check_agreement.py` |

`validate_release.sh` keeps its name and lives under `release/` because it is a release action, not a gate.

## Behavior-identical, with a before/after proof

This is moves plus a single entry point, not a rewrite. The only edits inside the moved files are mechanical: the repo-root path depth (`parents[1]` becomes `parents[2]` for the files now two levels deep, `parents[3]` for the tests three deep) and the relocated data-file paths. No gate logic was touched. `check_cli.py` and `check_python.py` are generated by the SDK-surface generator, so the generator's output paths and the validator templates were updated and the files regenerated to match; same for the one generated article that carries the snapshot provenance line.

Before any move, every gate, both generator `--check` runs, and the build were run on the pristine tree and their exit codes plus output recorded. After the move, the exact same command set was re-run through the new paths and diffed:

- Every gate's exit code matches the baseline exactly. The three gates that need a built FFI or a bench run (`check_c_abi_completeness`, `check_bench_regression`, `check_perf_gate`) fail identically before and after on a bare checkout; that is an environment precondition, not a logic change.
- The only output differences are non-behavioral: `check_docs_consistency` and `check_tier_badges` print the snapshot's new path in a status line, and the agreement test log differs only by its wall-clock duration (34 tests pass in both).
- `generate_sdk_surfaces --check` and `generate_docs_site --check` both still pass, which proves the build finds `upstream_openapi.yaml` at its new `scripts/ci/data/` location.
- `cargo build -p thetadatadx --locked` and `cargo fmt --all -- --check` pass.
- A repo-wide grep for `scripts/<oldname>` references finds zero outside the changelog and migration history.

Every workflow job, the build code that reads the snapshot and emits the validators, `.gitattributes`, and the in-tree references were repointed at the new paths.


userFRM:
Reopening on a clean branch name (scripts/ layout PR) — see the replacement PR.

PR #925, opened 2026-06-19
chore(release): prepare 13.0.0-rc.5

userFRM:
Bumps every workspace manifest and lockfile from 13.0.0-rc.4 to 13.0.0-rc.5 and regenerates the napi binding version guard in the TypeScript index.js so it reads rc.5. rc.4 (API-key authentication) is already out, and this records the one feature since then.

What shipped since rc.4:

Staging-environment selection (production or staging) across authentication, historical, and streaming, working with both API-key and email/password credentials. You can pick it in code via Config.stage() / with_environment(Environment::Stage), through the THETADATA_MDDS_TYPE environment variable (PROD or STAGE), or from a .env file via Config::from_dotenv(path) so one file can hold both THETADATA_API_KEY and THETADATA_MDDS_TYPE.

First-class inline client construction that takes the API key as a directly-passed argument: a fluent Client::builder() in Rust and C++, Python Client(api_key=..., mdds_type=...) plus Client.from_env(...) / Client.from_dotenv(...), and TypeScript Client.connectWith({ apiKey, ... }). The typed Client::connect(&Credentials, Config) remains for full control. api_key_from_env is strict in every binding, and the docs now lead with passing the API key directly.

This PR is release prep only: version bump, the new [13.0.0-rc.5] CHANGELOG section (kept byte-identical with the docs-site copy), and the v13.0.0-rc.5 release notes. No tag is cut here.

### DIFF
--- a/.github/release-notes/v13.0.0-rc.5.md
+++ b/.github/release-notes/v13.0.0-rc.5.md
@@ -1,0 +1,21 @@
+## ThetaDataDx 13.0.0-rc.5
+
+This release candidate adds staging-environment selection and first-class inline client construction. The only feature since rc.4; everything else is unchanged. See the [CHANGELOG](https://github.com/userFRM/ThetaDataDx/blob/v13.0.0-rc.5/CHANGELOG.md) for the full, itemized list.
+
+### Highlights
+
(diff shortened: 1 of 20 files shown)


userFRM:
Superseded: re-cutting rc.5 prep on top of the #926 hardening so the changelog covers the full rc.5 delta.

PR #1061, opened 2026-07-01
revert: expose contract_id on streaming data events (#1060)

userFRM:
Reverts #1060.

Keeping `contract_id` internal. It is a session-scoped wire routing key, not durable contract identity, and exposing it on the public event surface risks consumers treating it as a persistent identifier. Contract identity stays `(symbol, expiration, strike, right)`.


userFRM:
Superseded: removed contract_id via a clean reset of main to fba43d3c rather than a revert commit, so no revert is needed.

PR #1063, opened 2026-07-01
fix(fpss): resolve production-readiness audit findings (HIGH teardown, #1047, terminal parity)

### LINKED ISSUE #1047: fix(fpss): trades with 7 data fields are dropped for some contracts
## Problem

On the live stream, trade frames for some contracts decode to 7 data fields. The Trade arm expects 8 and marks anything else unparseable, so those contracts lose trade data. Because the decoder caches the field width from the first row per contract, once a 7-field row is seen first, later rows for that contract also read as 7 and drop. This has been observed on specific contracts while others decode normally at 8.

## Solution

Identify the 7-field shape (a distinct trade variant, for example a contract type with no exchange field, versus a decode miscount). This is under investigation with the data provider. Once confirmed, decode it on its own layout keyed on the concrete field count rather than the message type alone, so it does not seed or poison the 8-field baseline.

## Impact

Restores dropped trades for the affected contracts once the shape is confirmed.

userFRM:
Fixes surfaced by the production-readiness audit. No behavior change on conforming data; hardens the hostile-input and teardown paths and removes an SDK-invented mechanism.

## Fixed

- **[HIGH] C++ standalone `StreamingClient` teardown hot-spin + credential-handle leak.** On the no-callback teardown path, `thetadatadx_streaming_await_drain` returns 0 on an empty drain barrier; the C++ dtor / `operator=` read that as "callback wedged" and entered a sleepless reclaim loop that pinned a CPU core for 300s and then leaked the credential-bearing handle. Gate the reclaim on an installed callback (no callback → no consumer thread → synchronous teardown), and give the reclaim loop its own sleep. Teardown now completes in microseconds.
- **Narrow first row no longer poisons the delta width cache (closes #1047).** A complete-but-narrow absolute row (e.g. a 7-field trade) is now rejected at the width guard BEFORE it seeds the cache, so a later correct row re-seeds cleanly. The absolute-row guard rejects any width other than the shape's exact field count, uniformly across Quote / Trade / OHLCVC / OpenInterest / MarketValue — matching the terminal, which validates every stream tick's exact length. Delta rows are unaffected (narrower partial deltas still apply onto the prior baseline).
- **Removed the SDK-invented `handshake_deadline`.** The terminal has no handshake wall-clock cap; login is now bounded solely by the socket read timeout. A mute peer hits the pre-header transient → reconnect; a partial-then-silent peer is bounded by the per-stall no-progress timeout.
- **[C++] `check_tick_array`** frees the FFI array on the converter-throw path via RAII.
- **[batch_reader]** re-store the async waker only when it will not already wake the current one.

## Verification

Core + ffi + cpp build; 976 lib tests + the fpss integration suites pass; C++ Catch2 72/72; `clippy --workspace --all-targets --locked` clean; rustfmt clean; codegen drift + cross-binding parity clean.

Closes #1047

### DIFF
--- a/thetadatadx-cpp/include/thetadatadx.hpp
+++ b/thetadatadx-cpp/include/thetadatadx.hpp
@@ -468,15 +468,20 @@ std::vector<T> to_vector(const T* data, size_t len) {
 // FFI call so a stale error from a prior call isn't misattributed.
 template<typename T, typename Arr, typename Convert, typename Free>
 std::vector<T> check_tick_array(Arr arr, Convert convert, Free free_fn) {
+    // RAII the FFI allocation so it is released on every exit path, including
+    // a throw from `convert` (only `std::bad_alloc`) — otherwise the array
+    // would leak on the throw path.
+    struct FreeGuard {
+        Arr arr;
+        Free& free_fn;
+        ~FreeGuard() { free_fn(arr); }
+    } guard{arr, free_fn};
     const std::string err = last_ffi_error_raw();
     if (!err.empty()) {
         const int32_t code = thetadatadx_last_error_code();
-        free_fn(arr);
         throw_for_code(code, err);
     }
-    auto result = convert(arr);
-    free_fn(arr);
-    return result;
+    return convert(arr);
 }
 
 // Pure converter (no free): the `Convert` callback for a `ThetaDataDxStringArray`.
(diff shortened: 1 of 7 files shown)


userFRM:
Superseded by a re-scoped PR from a clean branch.

PR #1078, opened 2026-07-02
fix(python): release the callback lock before connecting so close() cannot deadlock

userFRM:
## Problem

The Python binding could deadlock the whole interpreter when `close()` raced an in-flight `start_streaming`. `start_streaming` held the shared callback `Mutex` across the entire detached connect and handshake, still writing through the guard after releasing the GIL. `close()` parks on that same mutex WITH the GIL held. Interleave: the connecting thread holds the mutex and has released the GIL, `close()` takes the GIL and blocks on the mutex, the connecting thread tries to re-acquire the GIL to finish, and neither can proceed. Reachable from `Client` / `AsyncClient` / `HistoricalClient` `close()` / `__exit__` / `__aexit__`, and Ctrl+C could not break it.

## Fix

- **Deadlock (root cause).** Reserve the registration slot in an inner scope, store the handle, and DROP the guard before `py.detach`, clearing the slot on a handshake failure. This ports the TypeScript binding's already-correct reserve-then-clear pattern. The double-start rejection plus the core stop-generation guard make the wide critical section unnecessary. Fixed in the Python `start_streaming` template (the codegen SSOT) and the checked-in generated surface was regenerated.
- **Closed-session attribute proxy.** The context-managed session's `__getattr__` swallowed the unified client's "client is closed" error with `if let Ok(stream)` and fell through, so `session.subscribe(...)` after `close()` raised `AttributeError` instead of the uniform closed error. The Unified arm now propagates the `stream` getter error with `?`.
- **Self-call drain.** `close()` invoked from inside a streaming callback ran the drain wait on the dispatcher thread itself, so the drain flag could never flip until the callback returned, always timing out and warning about the caller's own frame. A new core `StreamSurface::current_thread_is_dispatcher()` (mirroring the existing teardown self-join guard) lets `close_impl` skip the drain on that self-call.

## Tests / gates

- Added offline source-pin regression tests for the lock-release ordering and the closed-session proxy, plus a live-gated behavioral assertion that `subscribe()` after `close()` raises the closed error. A threaded start/close race is not reproducible offline (constructing a unified client needs a real handshake, and without a server the connect fails instantly so the race window is ~0), so the ordering is pinned against the generated source.
- `cargo fmt --check`, `cargo clippy --all-targets -D warnings` (py) and `--lib` (core), codegen-drift `--check`, the Python crate's Rust tests (55 passed), the core client-module tests (27 passed), and offline pytest all pass. No version bump.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - **Windows server archive attaches to the release.** The Windows `thetadatadx-server` archive was built, but the upload step received a path the artifact action could not resolve on Windows, so the GitHub Release carried no Windows binary. The archive is now referenced by its native workspace path and attaches correctly.
+- **Python `close()` can no longer deadlock against an in-flight streaming start.** `start_streaming` now releases its callback lock before the blocking connect, so a concurrent `close()` / `__exit__` / `__aexit__` on the client cannot wedge the interpreter.
+- **Python `session.subscribe(...)` after `close()` raises the uniform closed error.** The context-managed session's attribute proxy now surfaces the "client is closed" error on a closed unified client instead of masking it as an `AttributeError`.
+- **Python `close()` called from inside a streaming callback returns promptly.** Closing from the dispatcher thread now skips the drain wait it could never observe, instead of burning the full drain timeout and emitting a spurious warning.
 
 ## [13.0.0-rc.12] - 2026-07-01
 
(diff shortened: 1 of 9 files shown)


userFRM:
Follow-up: fixed a TOCTOU the reserve-then-clear change introduced. Dropping the callback lock before the connect opens a window where a concurrent stop + restart replaces the reservation mid-connect; the failed start then cleared the slot unconditionally, wiping the newer callback and leaving a live session with no registration (`reconnect()` had nothing to re-register).

The callable is now held behind an inner `Arc`, and the failure/panic clear runs through a `CallbackReservation` drop guard that clears only when the slot still holds its own arc (`Arc::ptr_eq`). The guard covers both the `?` failure return and a panic before the connect completes, and is disarmed on success. Added an offline source-pin regression for the ownership-checked clear.

Parity note: the TypeScript `startStreaming` has the same unconditional slot clear on handshake failure / task panic (`*cb_guard = None`) after dropping the lock, so it carries the same TOCTOU. Not changed here to keep this PR Python-scoped; worth a matching `Arc::ptr_eq` guard on the TS binding in a follow-up.

PR #1113, opened 2026-07-03
fix(python): run *_stream_async chunk handler off the shared runtime workers

userFRM:
Problem: the generated Python async streaming terminals (`*_stream_async`) invoked the per-chunk handler inline under the GIL on the shared async runtime that also drives every other in-flight historical call on the process. A slow or blocking handler held one of those shared workers under the GIL, so the network tasks serving unrelated concurrent historical requests were starved. It is a throughput degradation, not a deadlock: one caller with a slow async streaming handler degrades latency for every other concurrent historical request on the same client. The synchronous `stream()` path is unaffected because it drives the handler on the caller own thread via `run_blocking`, never a shared worker.

Changes: wrap the per-chunk GIL work (`Python::attach` building the pylist and calling the handler) in `tokio::task::block_in_place`, in the async stream template only. `block_in_place` hands this worker other tasks to a sibling thread for the duration, keeping the runtime pool at capacity so a slow handler can no longer occupy a serving worker. The fix lands in the code generator (`write_async_stream_terminal`), so every `*_stream_async` endpoint gets it uniformly with zero per-endpoint cost; the generated files were regenerated and verify with zero drift. The synchronous template is deliberately left untouched.

The streamed future remains the directly-polled future handed to `future_into_py`, which wraps it in the cancellation guard, so dropping the awaitable still drops the in-flight stream (RST_STREAM). `block_in_place` runs its closure inline and returns before the stream driver polls the next chunk, so once-per-chunk in-order delivery and the "chunk freed before the next is fetched" backpressure are unchanged. The captured-PyErr fold into the post-await converter (and thus the exception surfacing) is byte-identical apart from indentation.

Verification: `cargo run -p thetadatadx --features config-file --bin generate_sdk_surfaces --locked -- --check` exits 0 (zero drift). `cargo fmt --all --check` clean. `cargo clippy --manifest-path thetadatadx-py/Cargo.toml --all-targets --locked -- -D warnings -A clippy::too_many_arguments` clean. `cargo clippy --workspace --all-targets --locked -- -D warnings` clean. `cargo check --manifest-path thetadatadx-py/Cargo.toml` clean. The Python binding test suite passes (55/55), including the async-runtime concurrency overlap tests. `scripts/ci/check_binding_parity.py` clean (33 historical-streaming rows). `scripts/ci/check_docs_consistency.py` ok after copying CHANGELOG to the docs site. `scripts/ci/check_python.py` is a live production validator (imports the built wheel and connects to prod), so it is not runnable under the offline constraint of this change; the observable throughput improvement under concurrent live streams likewise needs a live server to measure, but the structural guarantee that the handler no longer holds a serving worker is correct by construction.


userFRM:
Superseded by #1116, which offloads the per-chunk handler to the dedicated blocking pool via spawn_blocking (through a new async for_each_chunk_async) instead of block_in_place. spawn_blocking keeps the shared runtime workers free for the HTTP/2 tasks and preserves drop-cancellation at an await point, so a cancel drops the stream immediately rather than waiting out an in-flight handler; it also avoids the current-thread-runtime panic footgun. Same observable fix, correct primitive.

PR #1159, opened 2026-07-04
chore(release): 13.0.0-rc.15

userFRM:
Version bump 13.0.0-rc.14 to 13.0.0-rc.15 across every pin (Cargo manifests + lockfiles, npm package.json, doc install snippets, OpenAPI version, READMEs) via scripts/release/bump_version.py plus the three prose pins the script's version= regex does not cover, with the CHANGELOG rc.15 section and its docs-site mirror. Version-sync and docs-consistency pass. Do not merge or tag until the release is approved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-## [13.0.0-rc.14] - 2026-07-03
+## [13.0.0-rc.15] - 2026-07-04
 
 ### Added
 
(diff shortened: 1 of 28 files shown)


CHANGELOG.md
```diff
@@ -7,11 +7,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- **The Python standalone streaming client supports the context-manager protocol without blocking the event loop.** `StreamingClient` gains `with` and `async with` support that stops the stream and drains on scope exit, mirroring the TypeScript client. The async exit runs the drain on a worker thread rather than inline, so `async with` teardown no longer parks the asyncio event loop while a slow callback drains.
+- **Streaming reconnect honors a server-sent rate-limit or restart signal.** An in-session `DISCONNECTED` reason from the server (too-many-requests, server-restarting) now reaches the reconnect classifier, so the client waits out the rate-limit cooldown instead of redialing on the fast transient ladder, uses the correct reconnect budget for a server restart, and no longer sits half-open after a transient disconnect the peer did not close.
```
userFRM, line 23: Does this match the terminal behavior ? Overkill ?

userFRM: This matches the terminal; it is not behavior the SDK invents. The reconnect classifier maps each server disconnect reason to a delay: a too-many-requests reason waits the rate-limit cooldown, an ordinary transient waits the standard delay, and a permanent reason (bad credentials, account-already-connected) does not reconnect. Those two delays are the wire-constant defaults the terminal itself uses.

The bug was that the decoded disconnect reason never reached that classifier, so the client ignored a too-many-requests signal and redialed on the fast transient ladder, which is the divergence from the terminal this fixes. The classifier already existed; this change just routes the reason into it.

The one intentional difference from upstream, documented at the classifier, is treating every credential error as permanent rather than only account-already-connected, so a bad credential no longer burns the rate limit. Net: terminal-matching, not overkill.

PR #1188, opened 2026-07-08
docs(fpss): describe flush_mode as outbound-only, not a data-latency tradeoff

userFRM:
flush_mode was framed as a "write-path latency vs syscall volume" tradeoff, which reads as if the default Batched mode delays received market data. It does not.

The setting gates only the client's **outbound** control frames — subscribe, unsubscribe, and heartbeat pings — at the two `IoCommand::WriteFrame` sites in the io_loop. The inbound read path dispatches trades, quotes, and OHLCVC continuously regardless of the mode, so flush_mode never adds latency to received data.

Batched (default) coalesces outbound frames and flushes them on the ping cadence (~100ms), so a subscription burst goes out as fewer, larger TCP segments — gentler on the server, fewer syscalls. Immediate flushes each frame so a subscribe reaches the server at once. Received-data latency is identical either way; the default stays Batched because coalescing outbound writes is the server-friendly choice.

Corrects the `StreamingFlushMode` enum doc, the io_loop flush-site comment, the Python/TypeScript setter docstrings (via `config_surface.toml` + regenerated accessors), and the streaming / configuration docs pages. Docs and comments only — no behavior change.


userFRM:
Superseded: flush_mode is an SDK-only invention (server-side write coalescing) with no equivalent in the Theta Terminal config surface — the terminal always coalesces at the socket and exposes no flush knob. Rather than document it, it will be removed from the client API (batching hardwired always-on) as part of the 0.2.0 parity pass. Closing in favor of the removal PR.

PR #1204, opened 2026-07-12
expose stream_window_size and connection_window_size for h2 (grpc)

aukaost:
## Summary

This change makes it possible for clients to adjust the above values to better utilize the h2 grpc connection.
The provided benchmarks show 40%-60% improvement depending on settings. 

## Changes
- exposes h2 stream_window_size and connection_window_size to client libraries
- raised default settings
- renamed existing `window_size` to `connection_window_size` for clarity

## Checklist

- [x] `cargo clippy --workspace -- -D warnings` passes
- [x] `cargo test --workspace` passes
- [x] Updated `CHANGELOG.md` (if user-facing change)
- [ ] Updated documentation (if public API changed)
- [x] FFI/SDK bindings updated (if core API changed)

## Testing
Provided examples for Rust, Python and JS
`cargo run --release --example chain_quote_bench`
`python examples/chain_quote_bench.py`
`npx tsx examples/chain_quote_bench.ts`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - **`ping_interval_ms` default is now `100`, matching the terminal.** The client heartbeat previously defaulted to `250 ms`; the Theta Terminal pings on a fixed `100 ms` period, and with `flush_mode` removed the ping is the sole flush trigger for queued outbound control frames, so the default now matches the terminal exactly (subscribe / unsubscribe reach the server within one `100 ms` interval instead of `250 ms`). The knob and its `[100, 300_000]` range are unchanged; override it if you prefer the old cadence.
- 
-- **exposed `stream_window_size` and `connection_window_size` to client libraries and raised default window sizes
+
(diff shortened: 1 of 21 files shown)


userFRM (commented):
Nice change, the core is solid: the clamp ceiling is exactly right (2_097_151 KB is the largest whole-KB value under the h2 2^31-1 byte cap), the KB→bytes conversion is safe since validate() clamps before the u32 cast, and the cross-binding surface (ffi/py/ts/cpp accessors, C header, .pyi, index.d.ts, parity.toml) all lines up with config_surface.toml. Codegen and parity checks pass.

A few things before merge:

1. `CHANGELOG.md:13-14` breaks CI and is malformed. Line 13 has trailing whitespace so `git diff --check` fails. Line 14 has an unclosed `**` (renders as literal asterisks) and uses non-canonical names (`stream_window_size` / `connection_window_size` instead of the `_kb`-suffixed fields). Same edits in `docs-site/docs/changelog.md`.

2. The rename is breaking and needs a migration note. `grpc.window_size_kb` becomes `grpc.stream_window_size_kb`, and `GrpcSection` is `deny_unknown_fields`, so an existing config.toml with the old key now fails to load rather than being ignored. Correct behaviour, but worth calling out, the `flush_mode` entry just below is the house pattern.

3. `thetadatadx-rs/src/config/mod.rs:752` rustdoc still says the window clamps to `[64, 1024]` KB. The impl clamps to `[64, 2_097_151]` at 996-1001 now.

4. `thetadatadx-rs/examples/chain_quote_bench.rs:3` names the internal transport in the doc comment. Your Python and TS ports of the same bench say "streaming market-data endpoint" (e.g. `chain_quote_bench.py:3`), match that wording here. The client-facing vocab gate doesn't sweep the rs examples directory, so adding `thetadatadx-rs/examples` to that glob would catch it next time.

5. Defaults: `stream=1024` / `connection=8192`. The pool opens N independent channels with roughly one in-flight stream each, so per-request rate is bounded by the stream window, not the connection window. `stream=1024` caps every request including plain single fetches, while the larger connection window mostly sits idle on a one-stream channel. Suggest `stream=8192` / `connection=16384`: the stream default is where the throughput actually comes from, and 8 MiB is a receive ceiling buffered as data arrives, not a fixed allocation. Your own bench ships tuned to 8192/32768, which makes the same point.

Also needs a rebase onto main, the red checks are the stale-branch conflict with #1203, not this code.

PR #1205, opened 2026-07-14
feat(mdds): config-driven bulk-fetch sharding for large history pulls

userFRM:
## Summary

Adds automatic, config-driven sharding of large buffered MDDS history pulls. With `bulk_fetch = Auto` (the default), a single large query is split into N balanced disjoint sub-requests, run concurrently across the tier's request budget, and merged back into one result — the caller writes one sequential query and the engine runs it at tier concurrency.

## Mechanism

A cheap density probe (coarse OHLCVC counts) estimates work along a split axis, N-1 cuts are placed at equal-cumulative-work quantiles, and N sub-requests fetch concurrently — each acquiring its own request-semaphore permit while the joining driver holds none, so the fan-out cannot deadlock the tier-sized pool. The split axis follows the query shape: time-bands for an intraday window, date-bands for a multi-day range. A size guard keeps small pulls, coarse-interval and end-of-day endpoints, and probe failures on the single-stream path, byte-for-byte unchanged.

## Output ordering

Concrete single-contract, stock, and index responses are byte-identical to the single stream — the merge is a plain in-order concatenation. Full-chain responses are merged into a deterministic canonical order (expiration, strike, right, ascending, and time-ascending within a contract) which is row-exact but not the server's arbitrary contract enumeration; `bulk_fetch = Off` returns the exact server order.

## Configuration

`market_data.bulk_fetch` (`auto` default / `off`) and `market_data.shard_concurrency` (defaults to the tier's request budget, clamped to the pool size) are exposed across all bindings. `MarketDataClient::bulk_fetch_plan`, with the public `ShardPlan` / `ShardBand` / `ShardQuery`, exposes the same plans for callers that prefer to drive concurrency themselves.

## Scope

Buffered (`.await`) history endpoints that reach tick-level row volumes benefit: option `trade` / `quote` / `trade_quote` and the greeks surface, stock `trade` / `quote` / `trade_quote`, and index `price`. The speedup scales with the account's request budget and is bounded by the server's per-account send rate. Streaming (`.stream`) sharding is a scoped follow-up.

## Verification

Row- and order-exactness were validated live against the single-stream path. Unit tests cover cut placement (including empty and single-bin histograms), axis selection, the size guard, empty-band handling (an empty sub-band folds to zero rows rather than failing the pull), and the merge ordering.


userFRM:
## Verified performance

Measured live on a Full-Pro account (8 concurrent-request slots) against a full-day SPXW 0DTE quote chain (`option_history_quote`, `strike=*`, `right=both`, `interval=tick`): 125,849,342 rows, ~15 GiB decoded. Same box, same day, same account for every row below. Ratios are the result; absolute times depend on tier, RTT, and the server's per-account send rate, so read them as this-box-this-day, not guarantees.

Two client-side levers stack. The progression below is streaming throughput (count only, apples-to-apples), each step measured:

| stage | full-day chain | vs old default |
|---|---:|---:|
| 64 KB window, single stream (the old default) | 872.7 s | 1x |
| 8 MB window, single stream (#1204) | 546.3 s | 1.60x |
| 8 MB window + 8-way sharding (this PR) | **155.0 s** | **5.63x** |

**Window (#1204)** raises the HTTP/2 flow-control window from the 64 KB protocol default to 8 MB, which stops the client flow-control-stalling the server. Its steady-state throughput gain is about 2.4x (18 to 43 MiB/s), but the full-day wall-clock gain is 1.60x (872.7 to 546.3 s) because the server's fixed response-prepare time does not scale with window size.

**Sharding (this PR)** then splits the one query into 8 balanced pieces across the tier budget and runs them concurrently, taking the already-windowed 546.3 s down to 155.0 s (3.52x on top of the window). Combined against the old 64 KB single-stream default, that is **5.63x** on Pro, and it scales down cleanly with tier as the concurrent-request budget drops (8 to 4 to 2 slots).

Every result is row-exact: 125,849,342 rows, verified against the single stream (byte-exact for concrete contracts; row-exact with a deterministic canonical order for chains).

Buffered `.list()` runs the same query in 377.7 s. It fans out the same 8 pieces (the fetch is the same ~155 s) but then builds the full 125M-row typed frame in memory and sorts it into the canonical chain order. That serial parse-and-sort is the difference; streaming does it incrementally as chunks arrive. A bounded-parallel post-processing pass (a follow-up) closes most of that gap. A single concrete contract (no chain, no sort) shards to 2.78x.

The remaining ceiling is server-side, not client: each shard still pays the server's response-prepare latency, and the whole account shares one per-account send rate, so the client can only fill the concurrency the tier already grants. Exact-row sharding (a server `offset`/`limit`), stream-as-generate instead of build-then-send, and cached immutable history would push this further, but those are server changes.

userFRM:
Closing to re-open with a single squashed commit and a clean history. Same branch, same change; new PR to follow.

userFRM:
Re-opened as #1206 (single squashed commit).

Issue #1027, opened 2026-06-29
Live-state tools in the MCP (materialized views over the stream)


userFRM:
## Idea

Give an LLM live access to the streaming feed without piping it the raw firehose, using the materialized-view-over-a-stream pattern: the stream runs continuously while the model cherry-picks the current state on demand.

## How it works

A background subscription keeps a queryable live state per watched contract, updated O(1) as each tick lands. The model does point-in-time pull queries against that pre-computed state and gets an answer in milliseconds, no matter how much data flowed. The stream is the river; the pull query is the cherry-pick. Prior art: ksqlDB pull queries over materialized views, Tinybird's MCP server, RisingWave stateful stream processing.

## Bundle into the existing MCP, not a separate server

These are just more tools on `thetadatadx-mcp`: same package, same `npx -y thetadatadx-mcp` install, same credentials and contract identity. The server becomes a stateful long-lived process holding a background streaming subscription alongside the historical query tools. No second server, no second install.

Caveat: streaming uses the account's single FPSS session, so `watch` must be opt-in and the operator should know it consumes that session.

## Tool scope — raw primitives only, no pre-built analytics

Expose raw live state plus the minimum commodity shaping so the model gets a summary instead of a firehose:

- `watch(contract)` — start maintaining live state for a contract.
- `latest(contract)` — the latest tick.
- `bars(contract, interval)` — downsampled OHLCV (commodity aggregation; the minimum that avoids flooding the model).
- `window(contract, seconds)` — the raw ticks in a bounded window, for callers that compute their own shaping.

Explicitly NOT pre-built here: VWAP, order-flow imbalance, VPIN, Lee-Ready, Greeks, or any derived signal. Those are the valuable analytics and stay in the analytics layer (Kairos) or the caller's own code. The data plane serves raw live state; the signals are the moat and are not given away in the free MCP.

## Open design question

Where exactly is the line? `latest` and the raw `window` are unambiguously raw. `bars` (OHLCV) is borderline — basic, but still an aggregation. Decide whether even OHLCV bars belong here, or whether the MCP ships only `latest` + raw `window` and the caller does all shaping. The tradeoff: more shaping in the MCP is friendlier to the model but creeps toward giving away analytics.

userFRM:
## Proposed surface

[...]

PR #1240, opened 2026-09-15
feat(mcp): speak the 2026-07-28 protocol revision

userFRM:
The Model Context Protocol's current revision is [`2026-07-28`](https://modelcontextprotocol.io/specification/2026-07-28/changelog). The server spoke `2025-11-25` and `2024-11-05`. This adds the new revision and keeps both old ones working.

### What the revision changed

It removed protocol-level state. There is no `initialize` handshake and no session: a client declares the revision it speaks on **every request**, in `_meta`, and the server accepts or rejects each request independently.

That moves version negotiation from one place to the dispatch path, which is most of this diff.

### What this PR implements

- **`server/discover`.** Mandatory in this revision. One request returns the revisions the server speaks, its capabilities and its identity, so a client can select a revision up front instead of probing. On stdio there is no HTTP status code to drive fallback, so this is also how a modern client detects a legacy server.
- **Per-request revision checking.** `io.modelcontextprotocol/protocolVersion` in a request's `_meta` is validated before dispatch. A revision the server does not speak is refused with `-32022` and a `data.supportedVersions` list, so the client can retry without a second round trip to find out what is on offer. The code sits in the `-32020..=-32099` band the specification reserves for itself, not the implementation-defined band this server uses for its own errors.
- **`resultType` on results.** `"complete"` everywhere; this server has no multi-round-trip path.
- **`tools/list` is cacheable.** `ttlMs` of 60s — the advertised set changes exactly once per process, when the background connect lands and the account's subscription becomes known. `cacheScope` is `private`, because that set depends on the authenticated account and a shared intermediary must never serve one caller's list to another.

### Backward compatibility

`2025-11-25` and `2024-11-05` clients are unaffected. They send no revision in `_meta`, which is read as *no claim* rather than a bad one, and the `initialize` handshake answers exactly as before with the negotiated `protocolVersion` in its result. No tool, argument shape or returned row changes.

### Not in scope

`subscriptions/listen` replaces `resources/subscribe` in this revision. This server exposes tools only, no resources, so there is nothing to subscribe to yet. It becomes relevant with #1027.

### Tests

Six added, covering the pure helpers and the dispatch path: revision negotiation across all three revisions, `_meta` extraction treating absent/empty/unrelated as no claim, the `-32022` payload and its error-code band, `server/discover` returning every supported revision plus identity, a refused revision not answering the request, and `tools/list` carrying the cache fields. `cargo test --locked` passes 38 unit tests and the integration binary; `cargo clippy --all-targets --locked -- -D warnings` is clean.


userFRM:
Review follow-up — one blocking finding was correct and is fixed.

**Fixed: the error payload had the wrong field names.** The refusal carried `data.supportedVersions`. The revision defines `data.supported` alongside `data.requested`:

```json
"error": {
  "code": -32022,
  "message": "Unsupported protocol version",
  "data": { "supported": ["2026-07-28", "2025-11-25", "2024-11-05"], "requested": "2099-01-01" }
}
```

A client's retry path reads both by name, so the old shape refused the request without telling the client anything it could act on. That is worse than not implementing the check.

**Fixed: `initialize` fell back to the wrong revision.** A legacy client sending a missing, empty or unknown `protocolVersion` was answered with `2026-07-28` — a handshake answered by naming a handshake-less revision the caller provably cannot speak. It now falls back to `2025-11-25`, the newest revision that still handshakes. The supported list is now built from the two named constants instead of repeating them as literals, so the two cannot drift.

**Fixed: identity on `tools/call` results.** `_meta["io.modelcontextprotocol/serverInfo"]` is now on the tool-call result too, not only discovery and `tools/list`.

**Fixed: the tests would have passed against an empty payload.** That was the fair hit. The error test built its own payload and the dispatch test only checked that `data` existed. They now assert the serialised response field by field — code, message, `data.requested`, `data.supported` by value, and the request id — and discovery compares the revision list by value rather than by length. `tools/list` now checks the identity its name promised, and a new test covers `resultType` and identity on a tool-call result.

Verified the new assertion bites rather than trusting it: renaming `supported` back to `supportedVersions` in the production path fails `a_revision_this_server_does_not_speak_is_refused` and nothing else. Restored, `cargo test --locked` passes 39 unit tests plus the integration binary, `clippy --all-targets --locked -- -D warnings` clean.

PR #1249, opened 2026-09-15
fix(dev): make the live smoke runnable again, and cover the relocated endpoints

### LINKED ISSUE #1243: Re-verify the calendar and interest-rate endpoints against the live service
### Problem

This SDK calls the calendar endpoints (`on_date`, `open_today`, `year`) and the interest-rate history endpoint on the same historical service and method namespace as everything else. The terminal has since moved its own equivalents onto dedicated services, separate from the market-data surface it uses for trades, quotes and greeks.

A refactor inside the terminal is not proof that the service we call changed. But these are the only endpoints in our surface whose terminal-side counterpart was relocated, which makes them the ones most likely to drift without notice, and nothing in the suite would catch it until a user's call failed.

### Solution

Run the live validation against the account and confirm each of these endpoints still answers on the path the SDK uses today:

- calendar on-date, open-today, year
- interest-rate history

If any has moved, update the method path and add it to the live smoke so the next move is caught within a day rather than at a call site.

### Impact

Four endpoints returning an error at call time instead of data, with no warning beforehand. Cheap to check, and the check is worth keeping whatever the answer turns out to be.

userFRM:
Closes #1243.

`just smoke` could not complete on any leg. Four separate things were stale, each one hiding the next.

| Leg | What was wrong |
| --- | --- |
| CLI | Invoked a binary that stopped existing when the CLI tool was dropped in #1011. Every run aborted here, before a single check ran. |
| Python | Called `client.calendar_open_today()` and `client.shutdown()`. Endpoints hang off `client.market_data`, and teardown is `close()`. |
| Server | Polled `/v3/system/status` for JSON. The server serves the vendor terminal's three management routes and their bodies are bare text; its data routes default to the terminal's CSV. |
| MCP | Listed tools immediately after `initialize` and required the connected set, racing the background connect. |

### What changed

The CLI leg is removed rather than rebuilt — the Python, server and MCP legs already cover the same endpoints, and there is no CLI to test.

The Python leg is rewritten against `client.market_data` with `close()` teardown, and its checks are a table so adding one is a line.

The server leg reads `/v3/terminal/mdds/status` as text and asks its three data routes for `format=json`.

The MCP leg polls `tools/list` until the connected set appears. That is not a workaround: the server deliberately advertises only what it can serve without a client until the connect lands, so a client is expected to re-list.

### The coverage #1243 asked for

`calendar_on_date`, `calendar_year` and `interest_rate_history_eod` are now checked. These are the only endpoints in the surface whose terminal-side counterparts moved onto services of their own, so a relocation on the historical service shows up here instead of at a user's call site.

### The answer to #1243

All four answer on the paths the SDK uses today, against the live account:

```
calendar_open_today            OK   rows=1
calendar_on_date               OK   rows=1
calendar_year                  OK   rows=12
interest_rate_history_eod      OK   rows=2
```

No SDK change is needed. The terminal's internal reorganisation did not move the surface we call.

### Verified

```
python sdk smoke: ok
server smoke: ok
mcp smoke: ok
live smoke: ok
```


userFRM:
Both nice-to-haves applied in the follow-up commit.

`_run` existed only to invoke the CLI binary and had no caller left once that leg went, and the module docstring still advertised CLI coverage. Both gone. Dead code is not optional in this repo, so neither of these was really a nice-to-have.

Re-verified against the live account after the change:

```
python sdk smoke: ok
server smoke: ok
mcp smoke: ok
live smoke: ok
```

PR #1250, opened 2026-09-15
feat(mcp): live tape tools — one call answers what a model actually asked

### LINKED ISSUE #1027: Live-state tools in the MCP (materialized views over the stream)

### LINKED ISSUE #1251: Reconcile the MCP stream registry against the SDK's subscription tracker
### Problem

The MCP stream tools (#1250) treat `subscribe` and `unsubscribe` as authoritative. Neither is.

`subscribe` returns once the command is enqueued. The server can reject it afterwards — `MaxStreamsReached`, `InvalidPerms` — and the SDK drops the subscription when it does. The caller is holding a handle that looks healthy and receives nothing. `stream_status` reports a growing age, which is indistinguishable from a quiet contract.

`unsubscribe` can fail, and the SDK keeps the subscription when it does. By then the registry has already dropped the book, so nothing will ever try to close it again. The allowance is held for the life of the process with no record of who holds it.

Both are the same shape: the registry believes it knows what is open, and it is guessing.

### Solution

Reconcile against the SDK rather than trust the return value.

1. Read the SDK's own active-subscription set after a change and compare it to what the registry holds.
2. On a subscribe that did not land, fail the `stream_watch` call and release the handle, rather than returning one that cannot work.
3. On an unsubscribe that did not land, keep the book referenced so a later release or the idle sweep retries it, instead of forgetting it.
4. Have `stream_status` report the divergence when it exists, so a caller can tell a dead subscription from a quiet one without guessing from age.

### Impact

Today a rejected subscription is silent: the model asks for a contract, gets a handle, reads nothing, and has no way to learn why. A failed unsubscribe silently consumes part of the account's concurrent-stream allowance until the process restarts. Neither is visible in any output the server produces.

Found auditing #1250. The reference counting and the idle sweep are correct on their own terms; the gap is that the terms assume the feed did what it was asked.

userFRM:
Closes #1027.
Closes #1251.

A model cannot read a feed: ticks arrive orders of magnitude faster than tokens. The useful shape is not to forward the stream but to have been in the room: hold the subscription, do the work as the rows arrive, remember where the model last looked, and answer in one call.

### Five tools

| Tool | What it answers |
| --- | --- |
| `tape_read` | what is it now, and what happened since you last looked |
| `tape_prints` | each trade with the quote that stood before it |
| `tape_market` | the whole market at once, filtered and ranked as it arrives |
| `tape_list` | what am I holding, and is each one alive |
| `tape_stop` | stop watching this contract |

No handles. A book is `(contract, kind)`, opened on the first read. The contract is already in the transcript because the user said it, and nobody wants two subscriptions to one contract — a handle would add a failure mode (forgetting it) and no capability.

Named by domain rather than mechanism: a model reading `stream_` reaches for plumbing too early.

### The whole market is a standing selection, not a buffer to search

`tape_market` reads every print in the market over the vendor's bulk trade stream. A read that filtered and ranked a buffer would be ranking over whatever fraction of a second the buffer happened to hold and calling it a window, and at a real open that fraction is very small.

So the caller's filter and rank are installed on the book instead, and every print is offered to them on the way in. The server keeps the top rows continuously and a read takes what is already held and starts again. The answer is correct whatever the rate, a print that is not kept is never built, and `examined`, `matched` and `unranked` disclose the population the returned rows were chosen from.

Ranking and filtering use the vendor's own fields. Nothing is derived: no greeks, no implied volatility, no aggressor side, and no bars built here. The vendor's own bar is served as the vendor sent it.

### The default window is "since your last read"

The server owns the cursor, so a model never has to remember when it last asked, and `new_since_last_read` is on every read. "Last N" is the size of a buffer, not the shape of a question.

### A result is a summary, not a transcript

Two hundred JSON objects repeating eleven keys is several thousand tokens the model then has to eyeball. Instead: health, then now, then counts and selections over the window — first and last trade, the high and low each carried as the whole message so its condition code travels with it, the condition histogram, the bid and ask span — then a short tabular tail with columns declared once, rows as arrays, `date` once per result, times as an Eastern clock so the model does no arithmetic, `age_ms` on the newest row.

### Health is in every result, never a separate call

`age_ms` on what it returns, and `feed` state so a dead feed is distinguishable from a quiet contract, since those look identical from age alone on an illiquid option. The read itself restarts a dead feed, since there is no subscribe step left to do it.

On the per-contract reads, `covers_seconds` and `clipped` say how far back the rows held actually reach and whether the window asked for reached further. A cumulative drop counter is noise: on a liquid quote book it reads in the millions within minutes. "The buffer reaches back 14 s, your 60 s window is clipped" is signal.

`tape_market` holds no buffer to fall short, so it carries neither: a print it did not keep was never built, and the only way one can be missing is the feed's own discard, which every read reports.

### Where the line on analytics sits

The server may select, count, sort and compare the vendor's own fields within the messages it holds, and states the population every summary covers. It may not apply a rule the vendor did not define, apply a model, infer intent, or bring a second source.

So: no greeks, no implied volatility, no gamma exposure, no aggressor-side inference, and no bars built here — which prints count and how cancels are handled is the caller's choice. Volume is deliberately absent from the summary for that reason; the condition histogram and the tail let a caller build it under their own rules.

But the feed sends **the vendor's own OHLC** ahead of each trade, and that is served as it arrives, labelled as the vendor's, never reconstructed from trades. Refusing to serve a bar the vendor sent, because we will not build one, would be inconsistent: vendor-derived is product, ours is not. Same principle keeps `quotes_after` on `tape_prints`, opt-in and off by default so it does not triple the row count.

### Correctness

Recovers from a dead feed: the handler is no longer installed behind a one-shot guard, so an io-thread fault or an exhausted reconnect budget no longer leaves every tool silently dead until restart. Subscriptions are shared and only the last reader closes one. A quote completes every print still short of its two, not merely the newest. An expired book is buried before the read rather than revived by it.

### Tests

82 pass, 42 on the tape surface. Every behaviour has a test that fails if the logic breaks — each one was proven by mutating the production path, watching the named test fail, and restoring the file byte-identical.

Validated against the live feed during extended hours: SPY quote and trade books, real NBBO with `age_ms`, and a print carrying the quote that stood before it.

`tape_market` validated against the live stock tape the same way: a selection ranked by size examined 1,014 real prints over 25 seconds and returned the five largest in order, `age_ms` 45; replacing the selection returned the previous one's rows with `selected_by` naming it; and a rank on a field no print carries is refused with the fourteen that work.


userFRM:
Rewritten after review. The over-engineering was real and worse than the constant that caught your eye.

**The actual problem: I invented a parallel contract identity.** A `Spec` struct, a `SecType` enum, a `Kind` enum and a `Tick` type — all mirroring things `thetadatadx-rs` already exports. In this repository of all repositories, a second definition of a contract is the wrong instinct twice over: more code, and a competing answer to the question the SDK exists to answer.

Now the book is keyed on `(Contract, StreamMsgType)`. Both are SDK types, both already hash, and the ring stores `StreamData` as it arrives — so `age_ms` comes from the `received_at_ns` the decoder already stamps rather than a clock read of my own. `scope` collapsed into `sec_type`, which `Contract` already carries.

**`stream_bars` is gone.** Aggregation has condition, cancel and size rules. Choosing them here bakes in a policy that belongs to the caller, and bars built on my assumptions would not reconcile with bars built anywhere else. `stream_window` serves messages with their condition and exchange codes intact instead.

**Also deleted:** `DEFAULT_RING` and `DEFAULT_TTL` as public configurable knobs nothing configured, a one-element constant array standing in for a `matches!`, a `Scope` enum duplicating `SecType`, a `Stamped` wrapper around a timestamp the tick already carried, and a public `Inner`.

**879 lines, from 1582.** Same behaviour under test: seven tests covering reference-counted sharing and last-release close, the bounded ring and its drop count, the two-quotes-after rule on a print, the index kind table, idle collection, refusal of an uncovered contract, and no book invented for an unwatched one. Clippy clean.

userFRM:
Second pass found two real ones, both correctness rather than style.

**The idle sweep leaked subscriptions.** It dropped a timed-out handle's reference and discarded what that freed — nothing closed them on the feed. A forgotten handle held an allowance for the life of the process, and the next watch on that contract opened a *second* subscription while the first was still tracked. That is exactly the failure the reference counting exists to prevent, so the sweep was quietly defeating its own purpose. It now returns what it freed and the caller closes it before opening anything new. There is a regression test named for it.

**A handle was committed before its subscribe succeeded.** On failure the caller got a handle that receives nothing, and worse, it held a reference that would make the next watch on the same contract skip subscribing too — so the failure propagated to a caller who did nothing wrong. Rolled back now.

**An unsubscribe that failed was swallowed** while the response reported the subscription closed. Counted and named separately now; claiming a close that did not happen tells the caller an allowance came back when the feed still holds it.

47 tests, clippy clean. 923 lines.

userFRM:
Two more, and the first is the one that mattered.

### The feed-stalling bug was mine, in the place I claimed it could not be

The streaming dispatcher calls `ingest` on the same lock a tool call takes. A read held that lock **across the entire JSON response** — so serialising a few hundred rows stalled the dispatcher for the duration. The SDK is explicit that the handler runs on the dispatcher and a slow handler blocks the event ring. I wrote "a reader never blocks the feed" in the module docs and then wrote exactly that.

Each reader now copies what it needs under the lock and releases it before building anything. The critical section is a clone of the rows being returned, nothing more. The tool layer also no longer reaches into the registry's internals: `latest`, `window`, `prints` and `status` are methods returning owned values, and the tests exercise those rather than a private map.

### sonic-rs versus serde_json, measured rather than assumed

Same payload shape this server actually returns — a quote row with both sides, size, exchange and condition — on this box:

| rows | bytes | serialize | parse |
| --- | --- | --- | --- |
| 1 | 238 | sonic 0.73 µs, serde 0.66 µs — **serde 9% faster** | sonic 0.65 µs, serde 2.19 µs — sonic 3.4× |
| 200 | 36 KB | sonic 101 µs, serde 89 µs — **serde 12% faster** | sonic 84 µs, serde 538 µs — sonic 6.4× |
| 4096 | 767 KB | sonic 3.61 ms, serde 4.39 ms — sonic 1.22× | sonic 1.86 ms, serde 14.9 ms — sonic 8.0× |

So the common assumption is backwards for this workload. **sonic-rs is slower at serialising** until roughly four thousand rows, and serialising is what an MCP server spends its time on — requests are one short line, responses carry the data. Its large win is parsing, which here is a 238-byte request where both are under three microseconds.

It stays, because it is already the choice across `tools/server`, `tools/mcp` and `json_canon`, and splitting one module off would trade a consistent dependency for no measurable gain in either direction. But the reason is consistency, not speed, and anyone carrying "sonic is faster" as a general belief should know it does not hold below about four thousand rows on the serialise path.

userFRM:
Audit round. Six findings against the committed head; two were already fixed in the preceding commit, one is unreachable, three were real and are fixed here.

**A second trade starved the first of its quotes.** A quote completed only the most recent print. The feed gives each trade the *next two* quotes, so whenever a second trade arrived before the first had both of its own — which on a liquid contract is most prints — the first kept one quote forever. A quote now completes every print still short of two, which is also why one quote legitimately belongs to two tails.

**A read revived an expired handle.** `touch` marked the handle used *before* testing its age, and the idle sweep only runs when the map is mutated. So a handle long past its window renewed itself on the very call that should have buried it, and would do so indefinitely. Age is decided first now.

**Releasing one kind wiped the other's prints.** Independently confirmed; fixed in the previous commit by keying books and correlation off one contract-level map.

Both new tests were run against the old behaviour and fail there. The correlation test asserts quote *identity*, not count, so duplicating one quote twice would not satisfy it — that was a fair hit on the earlier tests, which would have passed against broken correlation.

Not fixed, and why:

- **Registry mutation and the feed command are not atomic.** Real under concurrent calls, unreachable here: the stdin loop awaits each request inline, so a mutation and the client call after it are never interleaved. Documented on `execute` rather than built around. If anyone ever spawns per request, `watch` and `release` need a second mutex held across the client call — and explicitly not the ingest lock, which the dispatcher takes.
- **A failed unsubscribe loses cleanup ownership**, and **a server-side subscribe rejection leaves a handle that receives nothing**. Both are real. Both need the registry reconciled against the SDK's own subscription tracker rather than trusting the return of an enqueue, which is a larger change than belongs in this commit. Filing separately rather than half-doing it here.

PR #1258, opened 2026-09-15
ci(dev): catch napi drift locally, and diff what the build actually emits

userFRM:
Three pull requests failed on the same thing today — #1216, #1223 and one before it — because the committed napi output did not match a fresh build. The gate suite never rebuilt the bindings, so it was invisible until CI every time, and the fix was paid per pull request instead of once.

### A local gate

`just check napi_drift` rebuilds and diffs.

It runs `npm ci` first. Without that the build uses whatever `node_modules` is on disk, and after working on another branch that is a different napi toolchain — which emits different output, so the gate reports drift that is not in the tree. That is not hypothetical: it is exactly what the first version of this gate did to me, reporting `main` as drifted when the only difference was my own environment.

It is deliberately **not** in `ci.py all`. It compiles the crate, and a suite people stop running is worse than a slower one. `SLOW_GATES` records that as a decision rather than leaving the omission to be rediscovered.

If `npm` is absent it exits 2 and says so, rather than exiting clean and being mistaken for a check that ran.

### A hole in the CI check itself

CI only diffed `index.d.ts`. Both files are generated and both are committed — and `index.js` is the one that actually moved when the napi CLI was bumped, 222 lines of loader. So the check missed the drift it exists to catch. It diffs both now.

### Verified

Fails on a drifted tree with the file names in the message; passes on a clean one; `ci.py all` skips it; `ci.py list` shows it.


userFRM:
Both findings were real and are fixed.

**The check could pass on a file that had gone missing.** `git diff` ignores untracked paths, so deleting a generated file from the commit let the build recreate it as untracked and the diff reported the tree as clean — at exactly the moment a published artefact had disappeared. Both paths must now be tracked before anything is compared.

That check also now runs **before** the install and the build. It costs nothing, and it fails on a state no rebuild can fix; having it last meant waiting minutes to learn something knowable in a second. That ordering is also why I could not verify the first version quickly, which was the tell.

**It compared against the index, not HEAD**, so a staged but uncommitted regeneration read as in sync — the exact state the check exists to catch. Fixed locally and in the workflow.

Verified by deleting `index.js` from the index: fails in under a second, names the file, and exits 1.

Issue #1294, opened 2026-09-21
fix(mcp): the print and whole-market reads name a loss with one word where a read names the kind


userFRM:
A per-contract read now answers with `incomplete_because`, a closed set naming which kind of loss happened. The print read and the whole-market read still answer with a bare `feed_interrupted` boolean, and the print read with a bare `clipped` beside it.

Same word, two meanings, on sibling tools. Worse, the print read's `feed_interrupted` is keyed on a counter that also counts a row the feed delivered and this server refused, and a row it could not place — the same mislabelling that was just fixed on the read, still present here.

This is not a rename. Neither path has the state to name a kind from: the print correlation tracks a hole by position rather than by date, and a whole-market buffer records no loss date at all, only a count. Both need the kind recorded where the loss is, the way a per-contract buffer now does.

Worth doing together so the three tools answer in one vocabulary, and worth not doing by halves — a tool that names two of five causes and calls the rest an interruption is worse than one that says only that something is missing.

userFRM:
Closing: this describes the streaming MCP tools, which are not part of the shipped surface.

They live on `wip/mcp-streaming-parked` and have never been merged to main. Nothing on main carries a whole-market read, a print read, or the disclosure vocabulary this refers to.

The reason they are parked is the substance of this issue: a surface that answers from data it holds rather than data it fetches only works if every answer is honest about its own limits, and these tools do not yet meet that standard. Reopening the work means reopening that question first, not filing against the parked code.

PR #1295, opened 2026-09-21
feat(mcp): the listing says whether to believe each buffer

### LINKED ISSUE #1286: feat(mcp): a call that says whether the next answer can be trusted
Nothing tells a caller that the market is closed, or that a stream went quiet four minutes ago. A caller has to infer liveness from emptiness, which is the one inference these tools exist to prevent: a quiet contract and a dead feed look identical from an empty answer.

Add a call that takes no arguments and reports the session and the state of what is held: which phase the session is in and when it ends, per stream whether it is receiving, how far back it is known to cover, what it has discarded since it opened, and the losses it knows about recently.

The listing says what is held. This says whether to believe it.

userFRM:
Closes #1286.

A listing showed what a buffer holds and how busy it has been. It could not say whether to trust any of it: a contract quiet for an hour and one that *missed* an hour hold the same nothing, and a count of rows cannot tell them apart. A caller had to infer liveness from emptiness, which is the one inference these tools exist to prevent.

Each holding now carries `watched_for_seconds` — how far back it is known to have seen everything — and `last_loss`, when it last lost rows and what the loss was, absent where it has lost none so a listing shows only the buffers with a history.

**One vocabulary.** `last_loss.kinds` are the same words a read gives in `incomplete_because`, produced by one shared function. A listing and a read can no longer describe the same event two different ways; a test asserts them equal on one buffer, and a mutation that gives the listing its own word fails it.

**Honest where it does not know.** A whole market keeps a count of its losses and neither a date nor a kind, so its entry carries no `last_loss` whether or not it has lost prints. The description says that outright rather than letting an absent field read as a clean history. Closing that gap is #1294.

## Two calls I made rather than ask

**No new tool.** `stream_list` already takes no arguments and is where a caller asks what this server holds. Whether to believe it is the same question and wants the same call; a sixth tool would have duplicated five of its fields to add two.

**No network request.** A call whose job is to say whether the next answer can be trusted is worth nothing if it cannot answer while the market-data channel is down. Which session the market is in needs a calendar round trip, so it stays with the calendar tool, which is what that tool is for.

## Checks

179 MCP tests, clippy clean at `-D warnings`, docs and vocabulary gates passing. The vocabulary-sharing mutation was run twice: the first attempt left `kind_names` unused and failed to compile under `-D dead-code`, proving nothing, so it was rewritten to keep the function called and then killed the test properly.


userFRM:
Folded into wip/mcp-streaming-parked with the rest of the streaming work. The streaming tools are withdrawn from main in #1296; this change is preserved there rather than merged.

Issue #1285, opened 2026-09-21
feat(mcp): answer a whole-market question about an interval that has already passed


userFRM:
A whole-market read installs a standing question. It answers what matched since the last read, and everything that did not match is discarded as it arrives. So only one question per security type is live at a time, asking a second one throws away the first one's population, and a question nobody registered in advance can never be answered.

A caller exploring a market wants three looks at the same minute: the largest prints, then the widest spreads, then which names were busy. Today the first look destroys the data the second one needed.

Give the whole-market read a window, the way a per-contract read already has one, answered from retained prints rather than from a filter registered in advance. Same parameters otherwise.

Worth stating up front:

- Retention is bounded and varies with rate. Coverage is reported per call and a window reaching past it comes back clipped, as elsewhere.
- Three fields change meaning. Two of them describe a standing question that no longer exists, and the count of prints examined becomes rows scanned.
- Still no derived analytics. Counts, sums, extrema and top-N over a stated population only.

userFRM:
Closing: this describes the streaming MCP tools, which are not part of the shipped surface.

They live on `wip/mcp-streaming-parked` and have never been merged to main. Nothing on main carries a whole-market read, a print read, or the disclosure vocabulary this refers to.

The reason they are parked is the substance of this issue: a surface that answers from data it holds rather than data it fetches only works if every answer is honest about its own limits, and these tools do not yet meet that standard. Reopening the work means reopening that question first, not filing against the parked code.

PR #1297, opened 2026-09-21
WIP: the streaming MCP tools, parked for review

userFRM:
**Draft. Not for merge.** This branch holds the streaming MCP surface removed from `main` in #1296, so the work stays reachable and reviewable rather than only in history.

## What is here

Five tools over the streaming feed: a per-contract read with a bounded wait, a paired trade-and-quote tape, a whole-market read that filters and ranks every print as it arrives, a listing, and a stop. Plus the honesty vocabulary built around them: named loss causes, coverage that distinguishes a young buffer from a hole, and an explicit statement that gaps upstream of this server are undetectable.

Includes the listing work from the closed #1295.

## Why it is parked, not merged

**The account-session hazard.** A stream subscription belongs to the account, not the connection. Two connections on one account, the second acting on shared subscription state: on dev the incumbent is disconnected with `AccountAlreadyConnected` (permanent, not retried by design) reproducibly; on production it was unaffected in the same test. A feature that sometimes and silently kills a user's production feed cannot ship on "sometimes", and the audience for a whole-market read is Pro accounts, which are the ones with a production feed.

**The value case is unproven.** A contract's current state is answered better by a snapshot. What has no snapshot equivalent is aggregate measurement over a population between two moments, and no observed demand supports it yet.

## What would unpark it

1. Establish the account-session semantics with the vendor and document them, or find an isolation that does not contend.
2. Evidence that anyone asks questions needing a count over a population, rather than a price.
3. Decide whether the per-contract read survives at all, given snapshots answer it in one call with no state.

Open questions already filed: #1285 retention, #1294 the sibling tools still naming a loss with one word.


userFRM:
Closing rather than merging. These tools are not part of the shipped surface and were never going to land as they stand: the three issues describing them (#1285, #1286, #1294) are closed for the same reason.

The reason they are parked is the substance of the work. A surface that answers from data it holds rather than data it fetches only works if every answer is honest about its own limits, and these do not meet that standard yet.

The branch `wip/mcp-streaming-parked` is untouched, so nothing is lost. Reopening the work means reopening that question first.

PR #1329, opened 2026-09-22
fix(streaming): refuse an index market value instead of serving noise

### LINKED ISSUE #1321: fix(fpss): index market value is read as a quote and never emits
## Problem

Market value for an index is a different shape from market value for anything else. The vendor's reference client routes an index contract to the trade baseline, reads the row as an eight-field trade, and emits a separate event; every other security type goes to the quote baseline and the eleven-field quote shape.

This decoder has no security-type branch. Every `MARKET_VALUE` frame is read with the quote field count, so an index market-value row fails the exact-width check on arrival. The subscription is accepted upstream and reported by `active_subscriptions()`, and the book stays silent for ever.

Sharing the quote baseline per contract (#1298) does not reach this: an index needs the trade baseline and the trade shape, and a payload distinct from the current market-value event.

## Solution

Resolve the contract's security type at decode time, as the reference client does, and route an index to the trade baseline and the trade field count. Emit the index market value as its own event, which needs a new variant in the streaming event schema and regeneration across the four bindings.

## Impact

`thetadatadx-rs/src/fpss/decode.rs`, `thetadatadx-rs/src/fpss/delta.rs`, `thetadatadx-rs/fpss_event_schema.toml`, and the Rust, Python, TypeScript and C++ surfaces.

userFRM:
The terminal publishes an index market value. It is not a market value.

## What the terminal actually does

For a stock or an option it derives the figure from the NBBO, deterministically, and this SDK reproduces it field for field:

For a stock or an option the value is a deterministic function of the book: the bid and ask nudged by the size imbalance, and their integer midpoint.

For an index there is no NBBO, so it does this instead:

For an index the reference client does not derive the value from a book at all: it takes the last trade price and adds a random offset of up to five cents, redrawn on every tick. Two runs over the same tick give different answers, so there is no calculation to reproduce.

The last trade price, plus a random offset of up to five cents, redrawn for every tick.

## So this is refused, not implemented

#1321 asked for the index shape to be decoded and emitted as its own event, on the assumption that the terminal's separate branch carried real data. It does not. Serving it would publish noise as vendor data, which is the one thing this repository will not do, and the price the noise is drawn around is already on the trade stream, which is where an index publishes its prints.

The subscription is rejected at the same boundary as the other streams the feed does not publish (#1325), naming the trade stream as what an index does offer. The generated index market-value reference page is withdrawn along with its spec, and the streaming guide says why.

The availability test gains the case, so an index market value that starts being accepted again fails it.

## Scope

Only the index case changes. Stock and option market value are untouched: their derivation is the terminal's own defined calculation, it is deterministic, and the field doc already discloses that the value is derived rather than sent.

## Verification

Core suite green across all eight targets, clippy clean at `-D warnings`, gate suite green, both generator drift checks clean after regenerating the sidebar and `llms.txt`.

Closes #1321


userFRM:
Superseded by the branch this PR was rebuilt on. The premise was wrong: the vendor does publish an index market value stream, documented with `ms_of_day`, `date` and `market_price`. The replacement serves that stream with the price exactly as the feed sends it.

